PR #2, opened 2017-08-21
Make installable

rod-glover:
Fix some mistakes made when migrating this code to here.

Changes the published name of the package from `dp` to ce-dataprep`. The actual module imported is still called `dp`.


rod-glover:
Installed this version from our pypi server, with success. All scripts can be run.

PR #9, opened 2017-09-19
Add downscaled and downscaled-climo updates files

rod-glover:


jameshiebert:
This has been sitting around for a while. Did you still want it merged?

rod-glover:
Yes, I'd like to hold onto it. In Dec I started on a [repo](https://github.com/pacificclimate/data-prep-actions) that brings together all the ephemeral scripts and files and records of activities like these, but got interrupted by other higher priority activities. This will do in the meantime, and when I get a chance I'll migrate this and a bunch of other stuff scattered about into that repo.

Thanks.

PR #15, opened 2017-10-03
Add custom function to normalize experiment id values

rod-glover:
Also add some tests of custom functions.


rod-glover:
Hmm. I think codeclimate has too low a threshold for cyclomatic complexity. Ignoring these issues.

PR #20, opened 2017-10-05
Improve logging

### LINKED ISSUE #19: update_metadata: Log more useful information
1. Log only actions that are actually executed. E.g., don't log a rename that is excluded because the old metadata doesn't exist.
1. Log result of expression evaluation.

rod-glover:
Resolves #19


rod-glover:
Also passes end-to-end test

PR #32, opened 2017-10-10
Add access to dependent variable names

### LINKED ISSUE #26: update_metadata: Add access to dependent variable names
Extend the global variables context with function `dependent_var_names(arg)`. 
Value: output of `CFDataset.dependent_var_names(arg)`.

rod-glover:
Resolves #26 

Plus a few other goodies from CFDataset: filepath, dimensions, variables

### DIFF
--- a/dp/update_metadata.py
+++ b/dp/update_metadata.py
@@ -4,8 +4,10 @@ import logging
 import re
 import sys
 
-import numpy as np
 import six
+import yaml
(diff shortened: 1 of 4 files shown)


rod-glover:
codeclimates cyclomatic complexity check seems too tight: It's hard to imagine how to express the function `process_updates` more simply -- breaking it would actually create greater complexity.

PR #33, opened 2017-10-10
Add target key expression evaluation

### LINKED ISSUE #27: update_metadata: Add expression evaluation for keys
Allow the key of an update specifier to be the value of an expression. The resulting key value is processed as if it were a literal.

YAML syntax:

```yaml
"=<expression>": ...
```

rod-glover:
Resolves #27


rod-glover:
LGTM :)

PR #40, opened 2017-10-12
update_metadata: Add variable renaming

### LINKED ISSUE #36: Create (or modify) script to rename variables
Motivation: https://github.com/pacificclimate/modelmeta/issues/46

Task: Script similar to (or extension of) `update_metadata` that can rename a variable in a NetCDF file. 

Definitely tending towards extending  `update_metadata`, which already contains 90% of the machinery necessary for a nice implementation of this. Should rename it to something like `update_netcdf`.

rod-glover:
Resolves #36 

`update_metadata` is now not an entirely correct name, unless you regard the names of variables as metadata, which in fact one might reasonably do. So leave the name as it is.

### DIFF
--- a/dp/update_metadata.py
+++ b/dp/update_metadata.py
@@ -88,13 +88,12 @@ def cell_methods_for_var(var_name):
 
 # Helper functions
 
-def string_starts_with(prefix, string):
-    return (isinstance(string, six.string_types)
-            and string.startswith(prefix))
+def is_string_and_starts_with(prefix, thing):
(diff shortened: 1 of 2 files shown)


jameshiebert:
This has been sitting around for a while. Did you still want it merged?

rod-glover:
Yes, please. If you wouldn't mind having a quick look over the code, that would be great.

jameshiebert:
Yeah, it looks good to me. Don't know that I have too much to say aside from the use of ``partial`` is kinda clever. :)

PR #43, opened 2018-01-26
Add ncWMS vector formatting script

corviday:
Adds a straightforward script that processes the routing matrix from a VIC parametrization netCDF file into a netCDF file that ncWMS2 can display as vector arrows. 

**Rationale**: I've found being able to look directly at the routing information valuable for developing and testing watershed-related calculations. We may or may not want to include this visualization in a user-oriented version of Climate Explorer, but it's been useful enough for development it seemed worth documenting and preserving the script to generate it.

vectors


jameshiebert:
LGTM!

PR #46, opened 2018-02-21
Generate monthly climos from monthly data

jameshiebert:
The previous code falsely assumed that one cannot generate a monthly
average from a monthly climatology. However, this script creates
*interannual* averages, so it will generally have multiple months from
which to create an average.

Make sure that the cdo ymonmean command gets run even for monthly
data.


rod-glover:
Heh, nice fix. Not related to that warning, then.

rod-glover:
LGTM

PR #55, opened 2019-02-14
Add support for degree day climatologies

corviday:
Two minor changes to allow computation of climatologies for degree-day variables:

1. Add the degree day variables to `supported_vars`
2. Add possible operations for seasonal datasets (overall means, seasonal means) to `ops_by_resolution`


rod-glover:
Do any tests need to be updated to cover the 'seasonal' case? 

Otherwise, LGTM.

corviday:
Hm. I think the script may need further changes for degree day data. Currently, it outputs the annual value as the _mean_ of the four seasonal values, but that's pretty meaningless. Annual value as the _sum_ of the four seasonal values is what we want. Going to close this PR.

PR #58, opened 2019-02-28
Fix unused tests

nikola-rados:
This fixes #54 by updating the metadata found in some of the test files.  Using the [update_metadata](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/scripts/update_metadata) script and the pcic metadata standard I was able to rename a number of globals in each of the files.

### YAML files
Since the metadata changes can't be easily viewed, I will provide the `.yaml` files that I used with `udpdate_metadata`.

#### For `tiny_hydromodel_gcm_climos.nc` and `tiny_hydromodel_gcm.nc`
```
global:
        downscaling__method: <-forcing_downscaling_method
        downscaling__method_id: <-forcing_downscaling_method_id
        downscaling__package_id: <-forcing_downscaling_package_id

        downscaling__GCM__experiment: <-forcing_driving_experiment
        downscaling__GCM__experiment_id: <-forcing_driving_experiment_id
        downscaling__GCM__initialization_method: <-forcing_driving_initialization_method
        downscaling__GCM__institute_id: <-forcing_driving_institute_id
        downscaling__GCM__institution: <-forcing_driving_institution
        downscaling__GCM__model_id: <-forcing_driving_model_id
        downscaling__GCM__physics_version: <-forcing_driving_physics_version
        downscaling__GCM__realization: <-forcing_driving_realization

        model_cal__contact: <-forcing_target_contact
        model_cal__dataset: <-forcing_target_dataset
        model_cal__dataset_id: <-forcing_target_dataset_id
        model_cal__institute_id: <-forcing_target_institute_id
        model_cal__institution: <-forcing_target_institution
        model_cal__references: <-forcing_target_references
        model_cal__version: <-forcing_target_version

        method_id: <-hydromodel_method_id
```

#### For `tiny_downscaled_pr.nc` and `tiny_downscaled_pr_packed.nc`
This was the only place where missing data was added as oppose to being renamed.
```
global:
        project_id: 'CMIP5'
```

#### For `tiny_downscaled_tasmax.nc`, `tiny_downscaled_tasmax_climos.nc`, `tiny_downscaled_pr.nc` and `tiny_downscaled_pr_packed.nc`
```
global:
  GCM__experiment: <-driving_experiment
  GCM__experiment_id: <-driving_experiment_id
  GCM__institution: <-driving_institution
  GCM__institute_id: <-driving_institute_id
  GCM__model_id: <-driving_model_id
  GCM__realization: <-driving_realization
[...]

PR #59, opened 2019-03-01
Feature standard deviation added

nikola-rados:
This PR extends `generate_climos` to allow for different cdo stat operations, in particular **standard deviation**.  The referencing of _means_ has been removed from variable names in favour of more generalized names.  Furthermore, docstrings and comments have been updated to reflect the addition of standard deviation.  The `cell_methods` attribute is now updated according to the [metadata conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html) for both mean and standard deviation.  Tests have been updated to include the `operation` parameter.

This PR relies on the successful implementation of #58 (which fixes a collection of tests used here).


dp/generate_climos.py
```diff
@@ -124,62 +128,62 @@ def create_climo_files(outdir, input_file, t_start, t_end,
...
+    def climo_outputs(time_resolution, operation):
         """Return a list of cdo operators that generate the desired climo outputs.
         Result depends on the time resolution of input file data - different operators are applied depending.
         If operators depend also on variable, then modify this function to depend on variable as well.
         """
         ops_by_resolution = {
-            'daily': ['ymonmean', 'yseasmean', 'timmean'],
-            'monthly': ['ymonmean', 'yseasmean', 'timmean'],
-            'yearly': ['timmean']
+            'daily': ['ymon' + operation, 'yseas' + operation, 'tim' + operation],
```
rod-glover: You should check that `operation `has a valid value, and raise an exception with an informative message if not. Otherwise you will get a much harder to understand error a little later on.

dp/generate_climos.py
```diff
@@ -366,10 +370,29 @@ def update_metadata_and_time_var(input_file, t_start, t_end, climo_filepath):
             raise ValueError('Expected climo file to contain # time values in {}, but found {}'
                              .format(num_times_to_interval_set.keys(), cf.time_var.size))
 
+
```
rod-glover: You should validate `operation` here too. 

More generally, each function that takes it as an argument needs to do this. It's a bit repetitive, but it catches errors where they occur (e.g., mangling or forgetting to pass an argument). Given that, you probably want to create a utility function `validate_operation` that does this for you, and call it in each place.

dp/generate_climos.py
```diff
@@ -366,10 +370,29 @@ def update_metadata_and_time_var(input_file, t_start, t_end, climo_filepath):
...
+        # Update cell_methods to reflect the operation being done to the data
+        cell_method_op = ''
+        if operation == 'std':
+            cell_method_op = 'standard_deviation'
+        elif operation == 'mean':
+            cell_method_op = operation
```
rod-glover: This is harder than necessary to understand, and without validation of the value of `operation` can result in an erroneous value for `cell_method_op`. 

Assuming `operation` is validated, the following code is clearer and more concise. Generally speaking, using hashes to map one set of values to another is the clearest and cleanest tool for the job, and should be used wherever feasible. 

```python
cell_method_op = {
   'std': 'standard_deviation',
   'mean': 'mean',
}[operation];
```

dp/generate_climos.py
rod-glover: Also please add documentation for argument `operation`, including valid value set.

dp/generate_climos.py
```diff
@@ -366,10 +370,29 @@ def update_metadata_and_time_var(input_file, t_start, t_end, climo_filepath):
...
+                cf.variables[key].cell_methods = cf.variables[key].cell_methods + ' time: {} over days'.format(cell_method_op)
+            except AttributeError as e:
+                # skip over vars that do not have cell_methods i.e. lat, lon
+                continue
+
         # Update frequency attribute to reflect that this is a climo file.
+        suffix = ''
+        if operation == 'std':
+            suffix = 'SD'
+
```
rod-glover: Again, this is a mapping operation, so use a hash map.

rod-glover: Also, I think we should start using the suffix 'Mean' explicitly here, instead of continuing the past practice of not stating the operation and assuming mean in that case. This will go into the PCIC Metadata Standard when I get to it. @jameshiebert do you agree?

scripts/generate_climos
```diff
@@ -56,6 +56,8 @@ if __name__ == '__main__':
     parser.add_argument('-i', '--split-intervals', type=strtobool, dest='split_intervals',
                         help='Generate a separate file for each climatological period')
     parser.add_argument('-o', '--outdir', required=True, help='Output folder')
+    parser.add_argument('-p', '--operation', required=True, help='Data operation for the file',
+                        choices=['mean', 'std'])
```
rod-glover: 👍

tests/test_create_climo_files.py
```diff
@@ -44,18 +44,18 @@ def basename_components(filepath):
     # Slightly tricky because variable names can contain underscores, which separate components.
     # We find the location f of the frequency component in the split and use it to assemble the components properly.
     pieces = os.path.basename(filepath).split('_')
-    f = next(i for i, piece in enumerate(pieces) if piece in 'msaClim saClim aClim sClim mClim'.split())
+    f = next(i for i, piece in enumerate(pieces) if piece in 'msaClim saClim aClim sClim mClim msaClimSD saClimSD aClimSD sClimSD mClimSD'.split())
```
rod-glover: If we go ahead with the suffix 'Mean`, you'll have to add those cases here.

tests/test_create_climo_files.py
```diff
@@ -135,39 +135,43 @@ def test_filenames(outdir, tiny_dataset, t_start, t_end, split_vars, split_inter
...
+            assert cf.is_multi_year
             # In Python2.7, datetime.datime.isoformat does not take params telling it how much precision to
             # provide in its output; standard requires 'seconds' precision, which means the first 19 characters.
             assert cf.climo_start_time == t_start.isoformat()[:19] + 'Z'
             assert cf.climo_end_time == t_end.isoformat()[:19] + 'Z'
             assert getattr(cf, 'climo_tracking_id', None) == getattr(tiny_dataset, 'tracking_id', None)
 
+    suffix = ''
+    if operation == 'std':
+        suffix = 'SD'
```
rod-glover: Mapping. Also consider suffix 'Mean'.

dp/generate_climos.py
jameshiebert: Agreed!

dp/generate_climos.py
jameshiebert: Yeah, I'd second that motion.

nikola-rados: Wasn't the purpose of leaving it to ensure backwards compatibility?  Do we have to go back and change all the files to have `xClimmean`?

rod-glover: No, the idea is to go forward with being explicit about mean (`xClimMean`, note capital M) from here on in. In the case the operator is not specified (`xClim`), it is assumed to be mean. With that convention, we don't have to change any existing files, but we are explicit with all files created from here on in. It's a common backwards-compatibility strategy.

nikola-rados:
@rod-glover I've made the changes! Please let me know if there is anything else you would like changed or if I missed something.  Thanks again for looking through the code.

dp/generate_climos.py
```diff
@@ -85,9 +87,13 @@ def create_climo_files(outdir, input_file, t_start, t_end,
     """
     logger.info('Generating climo period %s to %s', d2s(t_start), d2s(t_end))
 
-    if input_file.is_multi_year_mean:
+    if input_file.is_multi_year:
```
rod-glover: The exception raised below should either use a non-specific term like "climatologies" or specify the correct climatological operation, not always say "climatological means".

dp/generate_climos.py
rod-glover: Not seeing the validation of `operation` here yet. ??

dp/generate_climos.py
```diff
@@ -394,3 +423,18 @@ def update_metadata_and_time_var(input_file, t_start, t_end, climo_filepath):
...
+    operation it must be in the cdo table of statistical values.
+    """
+    supported_operations = {
+        'mean',
+        'std'
+    }
+    if operation in supported_operations:
+        return True
+    else:
+        return False
```
rod-glover: This could be simplified to 

```python
return operation in supported_operations
```
The next item overrides this, but it's worth pointing out that this is pretty verbose and unnecessary.

rod-glover: That aside, you don't appear to do anything other than raise an exception when the test fails. So raise that exception inside this function and [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) up your usage code.

tests/test_create_climo_files.py
```diff
@@ -260,6 +270,16 @@ def test_time_and_climo_bounds_vars(outdir, tiny_dataset, t_start, t_end, split_
...
+                'mClimMean': 12,
+                'sClimMean': 4,
+                'aClimMean': 1,
+                'saClimMean': 5,
+                'msaClimMean': 17,
+                'mClimSD': 12,
+                'sClimSD': 4,
+                'aClimSD': 1,
+                'saClimSD': 5,
+                'msaClimSD': 17,
```
rod-glover: :+1:

tests/test_create_climo_files.py
```diff
@@ -277,8 +297,8 @@ def test_time_and_climo_bounds_vars(outdir, tiny_dataset, t_start, t_end, split_
             def d2n(date):
                 return date2num(date, cf.time_var.units, cf.time_var.calendar)
 
-            # Test monthly mean timesteps and climo bounds
-            if cf.frequency in {'mClim', 'msaClim'}:
+            # Test monthly timesteps and climo bounds
+            if cf.frequency in {'mClim', 'msaClim', 'mClimSD', 'msaClimSD'}:
```
rod-glover: :+1:

tests/test_create_climo_files.py
```diff
@@ -287,29 +307,29 @@ def d2n(date):
                     assert cb[0] == d2n(datetime(t_start.year, month, 1))
                     assert cb[1] == d2n(datetime(t_end.year, month, 1) + relativedelta(months=1))
 
-            # Test seasonal mean timesteps and climo bounds
-            if cf.frequency in {'sClim', 'msaClim'}:
+            # Test seasonal timesteps and climo bounds
+            if cf.frequency in {'sClim', 'msaClim', 'sClimSD', 'msaClimSD'}:
```
rod-glover: :+1:

dp/generate_climos.py
nikola-rados: Right, I should have caught that my mistake!

dp/generate_climos.py
nikola-rados: Sorry, I do a check a little higher up in the outer method, I forgot to add it inside the nested method.

nikola-rados:
@rod-glover I hope its all good now!

dp/generate_climos.py
```diff
@@ -394,3 +419,17 @@ def update_metadata_and_time_var(input_file, t_start, t_end, climo_filepath):
...
+    """
+    Given an operation assert that it is supported.  In order to be a supported
+    operation it must be in the cdo table of statistical values.
+    """
+    supported_operations = {
+        'mean',
+        'std'
+    }
+    if operation not in supported_operations:
+        raise Exception('Unsupported variable: cant\'t yet process {}'
```
rod-glover: One tiny change more: unsupported operation, not variable.

rod-glover:
LGTM

PR #61, opened 2019-03-18
Generate climatologies for counted variables like degree days

### LINKED ISSUE #57: Degree day annual data should be the sum of degree day seasonal data
I generated degree data climatologies, but I think the usual climatology approach where an annual value is the _mean_ of the seasonal value is meaningless for an accumulative value like degree days. I think we want annual values to be the _sum_ of seasonal values. I need to update `generate_climos` and redo that data.

corviday:
This PR adds a "sum" operation to generate_climos. This operation is for variables that represent a count of things that happen over the course of a year, such as frost days, which counts how many days in a year have a temperature lower than 0, or cooling degree days, which counts the number of degrees per day the temperature is above 18C.

Climos generated from these variables should be summed within a year. The number of frost days for a particular year is the _sum_ of frost days for each month. But they should be averaged between years - the frost days for thirty year climatology should be calculated by taking the _mean_ of all years. If "sum" is passed as the argument to generate_climos, variables will be treated this way.

This PR also makes some minor changes to support degree days specifically: adds degree days to the supported variables list, and adds support for generating climatologies starting with seasonal-resolution data.

An extra test data file, tiny_cdd_seasonal, provides tests for counted variables and seasonal-resolution data.

A question for reviewers:
Is there a clearer word to use for the name of the operation argument? It's called "sum", but the generated climatology isn't actually the sum of all values. Values are summed within a year, and then averaged between years. I considered "meansum" but wasn't sure that was any clearer.

Resolves #57


corviday:
Hm, some of the output data looks a little weird. Stand by for a bit more testing.

dp/generate_climos.py
```diff
@@ -117,11 +117,13 @@ def create_climo_files(outdir, input_file, operation, t_start, t_end,
         'rx5dayETCCDI', 'sdiiETCCDI', 'suETCCDI', 'thresholds', 'tn10pETCCDI',
         'tn90pETCCDI', 'tnnETCCDI', 'tnxETCCDI', 'trETCCDI', 'tx10pETCCDI',
         'tx90pETCCDI', 'txnETCCDI', 'txxETCCDI', 'wsdiETCCDI',
+        # Degree Day variables
+        'cdd', 'fdd', 'gdd', 'hdd'
     }
 
     for variable in input_file.dependent_varnames():
         if variable not in supported_vars:
-            raise Exception("Unsupported variable: cant't yet process {}".format(variable))
```
nikola-rados: Nice catch, that was probably me :nauseated_face:

dp/generate_climos.py
```diff
@@ -373,10 +421,13 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
...
 
         # Update cell_methods to reflect the operation being done to the data
+        # For this and for the frequency attribute, the "sum" operation is a mean:
+        # the between-years mean of the within-years or within-seasons sum.
         validate_operation(operation)
         cell_method_op = {
             'std': 'standard_deviation',
-            'mean': 'mean'
+            'mean': 'mean',
+            'sum': 'mean'
```
nikola-rados: So I guess at the end of the day this is a`mean` and not a `sum`, however I do think the metadata should specify the difference between the `mean` of some data and the `mean` of the `sum` of some data.  It _should_ reflect the fact that we've essentially done 2 sets of operations to the dataset, so something like: `time: sum over days time: mean over days`?  We should have a look at the [CF conventions](http://cfconventions.org/cf-conventions/cf-conventions.html#climatological-statistics) and figure out what would fit best.

dp/generate_climos.py
```diff
@@ -389,7 +440,8 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
         # Update frequency attribute to reflect that this is a climo file.
         suffix = {
             'std': 'SD',
-            'mean': 'Mean'
+            'mean': 'Mean',
+            'sum': 'Mean'
```
nikola-rados: Same as above, I think we have to consider the multiple operations.  Mostly because `frequency` is used later on in `nchelpers` to create a `unique_id`.  If we have a degree day file that uses `-p sum` and another that uses `-p mean` they will produce identical `unique_ids` even though they are not the same.  Thus, the indexing script will just replace one file with the other, rather than keep both (I just had this issue with gridded climatologies :cry:).

Looking at the [PCIC metadata standard](https://pcic.uvic.ca/confluence/pages/viewpage.action?spaceKey=CSG&title=PCIC+metadata+standard+for+downscaled+data+and+hydrology+modelling+data) (in the **PCIC standard frequency attribute values** section), the `frequency` attribute has an `<operator>` associated with it.  So potentially that description could be extended to include multiple operations (i.e. `mClimSumMean`) but I think that is a slippery slope.  My gut says that the suffix should be `sum` however, that too could cause problems in the future (i.e.  if we **just** wanted to sum the data).

I think this requires some discussion, perhaps @rod-glover could help us here.

tests/test_create_climo_files.py
```diff
@@ -158,22 +162,46 @@ def test_climo_metadata(outdir, tiny_dataset, operation, t_start, t_end, split_v
...
         'std': 'SD',
-        'mean': 'Mean'
+        'mean': 'Mean',
+        'sum': 'Mean'
     }[operation]
 
     if split_intervals:
         assert frequencies == {
            'daily': {'mClim' + suffix, 'sClim' + suffix, 'aClim' + suffix},
            'monthly': {'sClim' + suffix, 'aClim' + suffix},
```
nikola-rados: This isn't something that you changed but shouldn't this be `'monthly': {'mClim' + suffix, 'aClim' + suffix}`?

tests/test_create_climo_files.py
```diff
@@ -158,22 +162,46 @@ def test_climo_metadata(outdir, tiny_dataset, operation, t_start, t_end, split_v
...
+           'seasonal': {'sClim' + suffix, 'aClim' + suffix},
            'yearly': {'aClim' + suffix},
         }[tiny_dataset.time_resolution]
     else:
         assert frequencies == {
             'daily': {'msaClim' + suffix},
             'monthly': {'saClim' + suffix},
```
nikola-rados: Same as above, should it be `'monthly': {'mClim' + suffix}`?

dp/generate_climos.py
corviday: I think the cell method for non-climatology degree days going into this script is already `time: sum`, because that's the definition of degree days, sum the number of degrees above each threshold each day tp get the degree_days in a season, month, or year. Though perhaps I should make it say 'time: sum over days` instead?

So I don't know if we need to add anything extra to cell_methods besides the usual climatology stuff.

dp/generate_climos.py
corviday: Hm. I guess I'm starting to feel like I took the wrong approach to this problem. This isn't really a "sum" operation, despite me making it attached to the "operation" flag, more like a mean that CDO isn't smart enough to calculate in a single step, that I borrowed the nice stdev operation infrastructure to add. Adequately effective, but not the right approach.

Maybe, like @rod-glover suggested in our meeting yesterday, it should be a `data-type` flag - data either represents a (possibly simulated) **measurement**, in which case yearly values are the means of daily values, or a **count** of the number of times an event occurs, in which yearly values are the sums of daily values. Measurement would be the default data type.

That gives me an extra headache, though: I would have to figure out how to make cdo calculate the standard deviation of  "count" data types. It would be a lot of extra code, but maybe code that better reflected the mathematical reality.

rod-glover: Well, here's a nice philosophical question: In which of the following ways do we regard the final product? Because that is what both `operation` and `frequency` signify.

1. It is the climatological mean of an abstract variable that spans years and that _just happens, as an insignificant computational detail_ to be computed from finer-resolution inputs along the way.
1. It is a more complex statistic that must fundamentally be regarded as the (mean | sum) within years followed by mean over years on the finer-resolution variables.

Let's consider the arguments for both cases in the context of count variables. 

1. The abstract variable is a count variable, which is necessarily counted (summed) over a specified period (e.g., day, month, season, year). In that case, "sum within _period_" is just part of the definition of the variable of interest, and climatology (mean over years) of the variable of interest is the operation we are actually interested in.
1. The abstract variable is still a count variable, but it is counted (summed) over a shorter period and must be further processed to reflect the real period of interest. The operation we are interested in is actually this process of both summing up to the period of interest and then forming climatologies.

(The arguments are the same for point/continuous variables, but the nuances get obscured by the commonality of the operation mean in both the within-period and across-years operations.)

I think case 2 is significantly weaker. In which case, this is just a (climatological) mean, and the grotty details of computation are not to be placed in the filename/unique id. And (phew) we do not need to consider using compound methods in the `frequency` attribute just yet. `xClimMean` will continue to suffice.

However, @nikola-rados , your observation about `-p sum` and `-p mean` operations yielding the same unique ids is important. It's important because `-p mean` (in the sense of mean within years) _is invalid_ for degree-day and other count variables. And therefore we need, at minimum, to add operation checking that ensures that invalid combinations of variables and operations aren't allowed to proceed. 

Even better, I think we should eliminate the "sum" operation, and always call it "mean" (signifying _climatological_ mean, which is actually the point of the computation). The _code_, not the user, decides how to compute that climatology, based, in part, on the kind of variable it is processing. In the case of point variables (e.g., temperature), it is mean of means; in the case of counts (e.g., gdd), it is mean of sums. This would eliminate the possibility of user-induced error. 

This will simplify some parts of the code while introducing some complications elsewhere. I believe those complications are quite justified, and less inclined to lead us into problems later.

To sum up, we need to distinguish between _climatological_ operation (mean or standard deviation) and intermediate computational operation (sum or mean, dependent on variable type) that achieves the desired end given our inputs. It so happens that one case is prepackaged for us by CDO, and the other must be specially coded (by Lee). This is a very desirable distinction to make, because one is specification (climatological operation), and the other is means/implementation. In general we want to highlight the former and hide the latter.

rod-glover: Aha, @corviday , I think you arrived at much the same conclusion as I did.

rod-glover: And you are right: Standard deviation of count data types may take some complicated code. But it's very worthwhile to consider that case, because it foregrounds the distinctions that got lost in case of mean.

Questions:

1. What actually are we taking the standard deviation of in the case of point/continuous variables (e.g., temperature)? Is it:
   1. SD(means of observations within years), taken across years
   1. SD(all observations)

1. What are we taking standard deviations of in the case of count variables (e.g., gdd)? Is it:
   1. SD(sums of observations within years), taken across years
   1. SD(all observations)

For both types of variables, the two different SD's are mathematically different, i.e., give different values. This is unlike the case for means, where the formulas are essentially the same.

corviday:
A list of variables that are counts:

* All degree_day variables:
  * cdd
  * gdd
  * fdd
  * hdd
* The following climdex variables which count the number of days (or nights) something happens: 
  * fdETCCDI (frost days)
  * suETCCDI (summer days)
  * idETCCDI (icing days)
  * trETCCDI (tropical nights)
  * r10mmETCCDI (days with more than 10mm precipitation)
  * r20mmETCCDI (days with more than 20mm precipitation)

The following variables represent a new headache: climdex variables that measure the duration of a continuous sub-year span of time, and sort of aren't either count variables _or_ measurement variables:
  * gslETCCDI (growing season length) 
  * wsdiETCCDI (warm spell duration index)
  * csdiETCCDI (cold spell duration index)
  * cddETCCDI (dry spell length)
  * cwdETCCDI (wet spell length)

You cannot generate yearly data from seasonal data of these variables without additional information. For example, if you have a wet spell duration from spring and a wet spell duration from summer of some year, you would want to add their lengths together if the spring one goes to the last day of spring and the summer one begins on the first day of summer; they'd be one continuous wet spell. But if they don't overlap temporally, the value for the year would be whichever wet spell was longer. The dataset doesn't tell you when the wet spell is, only how long it is.

Thankfully, we've never encountered and probably never will encounter seasonal data for any of these indexes. Should generate_climos just throw an error message if it does?

tests/test_create_climo_files.py
rod-glover: Hmm. When we have an input dataset with a time resolution of <= monthly, then we form monthly, seasonal, and annual climo means, yes? Which means that the monthly case should be `'monthly': {'mClim' + suffix, 'sClim' + suffix, 'aClim' + suffix}`.

tests/test_create_climo_files.py
rod-glover: Same as above: Nope, it should be 'monthly': {'msaClim' + suffix}, for the same regions. Same question about testing.

tests/test_create_climo_files.py
rod-glover: The code that generates it is as follows, and I believe is correct. So this is a testing issue, not a script issue.

```
prefix = ''.join(abbr for interval, abbr in (('monthly', 'm'), ('seasonal', 's'), ('annual', 'a'), )
                         if interval in interval_set)
        cf.frequency = prefix + 'Clim' + suffix
```

where `interval_set` is a subset of {'monthly', 'seasonal', 'annual'} as determined by the file's actual contents.

corviday:
I think this is ready for another round of reviews. 

Now, when someone runs `generate_climos` on a dataset, the only valid operations are `mean` and `std`. The script checks the variables in the incoming dataset against a giant list it has of variables types. In all cases, it generates climatologies with means or standard deviations of the same time resolution as the variable, but depending on the variable type, it may also:

- Generate additional climatologies for other time resolutions using cdo commands in a single step
- Generate intermediate data files for other time resolutions and then generate climatologies from them, using cdo commands in 2 steps
- Not generate climatologies for any resolutions that don't match the incoming data

Also added new tests for the second two scenarios. (The first used to be the default, so there were already tests for it) and corrected the bug in tests @nikola-rados found earlier.

dp/generate_climos.py
rod-glover: @corviday , I am not sure what you mean here. CF Conventions is quite clear about what constitutes a valid climatology `cell_methods`. But I'm not sure what your comment implies about the `cell_methods` we should assign to these climatologies.

corviday: I was answering @nikola-rados, but we both posted simultaneously, or near enough I didn't see your comment first. Probably it's confusing because it wasn't intended to address anything you said.

I said that going into the climatology operation, degree_days had a cell_methods of `time: sum` and thus came out of it with cell methods `time: sum time: mean over days`.

tests/test_create_climo_files.py
```diff
@@ -158,23 +165,70 @@ def test_climo_metadata(outdir, tiny_dataset, operation, t_start, t_end, split_v
...
+            with CFDataset(cf) as out:
+                outvar = out.variables[var][:]
+                timeres_ordinality = {"daily": 1,
+                                      "monthly": 2,
+                                      "seasonal": 3,
+                                      "yearly": 4}
+                inres = timeres_ordinality[tiny_dataset.time_resolution]
+                outres = timeres_ordinality[out.time_resolution]
+                print(outvar)
+                if(outres > inres):
```
rod-glover: No parentheses around if conditions in Python.

dp/generate_climos.py
rod-glover: Regrettably, that isn't one of the valid values for `cell_methods` required by [CF conventions, sec 7.4 Climatological Statistics](http://cfconventions.org/cf-conventions/cf-conventions.html#climatological-statistics). You have to specify the summation period. `sum within <interval>` is correct, where `<interval>` is the climatological sub-year interval. `sum over days` isn't right; it isn't even in the right form.

dp/generate_climos.py
```diff
@@ -92,64 +93,152 @@ def create_climo_files(outdir, input_file, operation, t_start, t_end,
...
+        'tx90pETCCDI': 'point',
+        'txnETCCDI': 'point',
+        'txxETCCDI': 'point',
+        'wsdiETCCDI': 'duration',
+        # Degree Day Variables
+        'cdd': 'count',
+        'fdd': 'count',
+        'gdd': 'count',
+        'hdd': 'count',
+        }
```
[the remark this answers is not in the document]
nikola-rados: Agreed, great job here

dp/generate_climos.py
corviday: Per face to face discussion with @rod-glover, cell methods changes probably need to be a separate issue and a separate PR. I won't be addressing them here.

corviday:
I think I got all the requested changes, except for the discussion on cell_methods, which has been bumped to [another issue](https://github.com/pacificclimate/climate-explorer-data-prep/issues/64) and will be solved separately.

rod-glover:
Nice job, @corviday . This was tricky and important.

PR #63, opened 2019-03-22
Snowfall generation script

nikola-rados:
### Overview
This PR adds a new script that generates `snowfall_flux` with data from `precipitation_flux`, `tasmin` and `tasmax`.  In addition to the script some tests and test files have been added.

### `generate_prsn.py`
This script takes in three netcdf files (`pr`, `tasmin`, `tasmax`) and uses each of them to create the `prsn` output.  The mean of the two temperature files is used to determine whether or not a particular data cell had a temperature below freezing.  Where temperature is freezing the `pr` data is used, otherwise 0.  This is then output to a new netcdf file.

### Chunks
The main issue encountered while building this script was reading/writing extremely large files.  To solve this all three data file are read in chunks and the output file is written in chunks.  These chunks are constant along the `time` dimension such that we are reading 100 slices.  Therefore the chunk size is partially determined by how large the spatial dimensions of the file are.  I think that it would be a good idea to find a way to dynamically select the most efficient chunk size.  However, this method has been tested with the biggest files I could find and it works for now.

### Metadata Check
Running the script with the `-m` (`--metadata-check`) flag will display all the required metadata to run the script and the logger will complain if there is missing metadata.  This was originally part of the `preprocess_checks(...)` method, however I decided to move it out.  As I was testing different files I found myself using the method similarly to how [`generate_climos's --dry-run flag`](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/scripts/generate_climos#L11) is used and thus thought it would be a good idea to isolate it.  I considered moving out all of the pre-processing checks out to as well, but decided to leave them in `generate_prsn_files(...)` for now.


dp/generate_prsn.py
```diff
@@ -0,0 +1,272 @@
...
+       Parameters:
+            pr_filepath (str): The filepath to desired precipiation data
+            tasmin_filepath (str): The filepath to desired tasmin data
+            tasmax_filepath (str): The filepath to desired tasmax data
+            outdir (str): Output directory
+    '''
+    for filepath in [pr_filepath, tasmin_filepath, tasmax_filepath]:
+        logger.info('Retrieving file: {}'.format(filepath))
+
+    # datasets
```
jameshiebert: How about changing the filepaths parameter to be a hash. Then you can do something like:

```
datasets = { varname: CFDataset(filepath) for varname, filepath in filepaths.items() }
variables = { varname: datasets[varname].variables[varname] for varname in filepaths.keys() }
required_vars = filepaths.keys()
```

Simplifies a bit of the repetition maybe?

scripts/generate_prsn
```diff
@@ -0,0 +1,41 @@
...
+def main(args):
+    if args.metadata_check:
+        logger.info('Metadata Check')
+        for filepath in [args.prec, args.tasmin, args.tasmax]:
+            logger.info('')
+            logger.info('File: {}'.format(filepath))
+            try:
+                dataset = CFDataset(filepath)
+            except Exception as e:
+                logger.info('{}: {}'.format(e.__class__.__name__, e))
```
jameshiebert: Should this be `logger.exception`? A) then the exception info gets logged automatically, and B) this program can't continue anyways if you couldn't open one of those datasets. C) then you can drop the `else` clause, making the rest of that block a little more readable.

scripts/generate_prsn
```diff
@@ -0,0 +1,41 @@
...
+            else:
+                for attr in 'project model institute experiment ensemble_member'.split():
```
jameshiebert: Maybe split this out into a utility function. It's a little dense for the `main()` function.

README.md
```diff
@@ -187,14 +187,14 @@ attributes. There are no really convenient tools for updating metadata, so we ro
...
-`update_metadata` takes an option (`-u`) and an argument: 
+`update_metadata` takes an option (`-u`) and an argument:
 
 * `-u`: the filepath of an updates file that specifies what to do to the metdata it finds in the NetCDF file
 * argument: the filepath of a NetCDF file to update
 
 #### Updates file: specifying updates to make
 
-`update_metadata` can update the global attributes and/or the attributes of variables in a NetCDF file. 
+`update_metadata` can update the global attributes and/or the attributes of variables in a NetCDF file.
```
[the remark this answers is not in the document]
nikola-rados: Ah sorry about that, I think my IDE will fix things like that automatically.  I'll have a look and see if its a setting I can turn off!

dp/generate_prsn.py
```diff
@@ -0,0 +1,272 @@
...
+import os
+
+from statistics import mean
+from math import floor
+from nchelpers import CFDataset
+
+from dp.units_helpers import Unit
+
+
+# Set up logging
```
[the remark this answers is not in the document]
nikola-rados: Do you mean you want it similar to how it was done with [crmprtd](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/__init__.py#L138)?  Set it up with a yaml file and such?

dp/generate_prsn.py
```diff
@@ -0,0 +1,272 @@
...
+    '''
+    required = {
+        'project': [],
+        'model': [],
+        'institute': [],
+        'experiment': [],
+        'ensemble_member': []
+    }
+
+    for dataset in datasets:
```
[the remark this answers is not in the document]
nikola-rados: Sorry I'm not sure I understand.  This part doesn't have a `required` variable like the comment above, are you suggesting that I change the lists into sets?  Or something else?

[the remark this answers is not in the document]
jameshiebert: Oh I'm sorry. I commented a bit hastily and misunderstood what you were doing here. I just read the `required` variable and assumed you were doing something similar (ensuring required attributes were set).

Now that I've read it more carefully, I'd say that this definitely works. I'd lean toward writing it in a single loop (required attr in the outer loop, and then a list comprehension iterating over the datasets in the inner loop). But... this works, so don't worry about it.

dp/generate_prsn.py
```diff
@@ -0,0 +1,272 @@
...
+        for attr in required.keys():
+            required[attr].append(getattr(dataset.metadata, attr))
```
rod-glover: Nope, I think James was right. Comprehensions are to be preferred over explicit loops. Easier to read and understand, and less room for error:
```
required_attrs = {'project', 'model', 'institute', 'experiment', 'ensemble_member'}
required = { 
   attr: [getattr(dataset.metadata, attr) for dataset in datasets] 
      for attr in required_attrs
} 
```
You might want to eliminate the intermediate variable `required_attrs`, though it does provide some explanation by its name.

rod-glover:
Regarding the metadata check, we should consider whether it belongs in [nclint](https://github.com/pacificclimate/nclint). @jameshiebert ?

nikola-rados:
I've made a number of changes to address the feedback from both @rod-glover and @jameshiebert .  I've also done a fair bit of testing to ensure that the script works as advertised.  I think this is ready to go! Let me know if you guys have any other comments!

jameshiebert:
Yeah, this is much better. Thanks for improving it! I'm happy w/ it if you are.

nikola-rados:
Sorry @rod-glover I totally whiffed on a couple of your suggestions.  I'll work them in!

nikola-rados:
If you are happy with it now @rod-glover I'd be happy to merge this!

dp/generate_prsn.py
```diff
@@ -0,0 +1,278 @@
...
+
+def create_filepath_from_source(source, new_var, outdir):
+    '''Using the source cmor_filename build a new output filepath with new_var
+       and output directory.
+    '''
+    variable, *rest = source.cmor_filename.split('_')
+
+    suffix = ''
+    for var in rest:
+        suffix += '_' + var
```
rod-glover: Sorry I didn't spot this before: Why not use `suffix = '_' + '_'.join(vars)`? One line, clearer. Your call.

rod-glover:
One tiny tweak, otherwise good to go! Nice work, @nikola-rados !

P.S. You may have to get @jameshiebert to approve his review on GitHub before it will let you merge.

dp/generate_prsn.py
nikola-rados: Your way does look much better!

PR #66, opened 2019-03-29
Add prsn support

### LINKED ISSUE #65: Generate Climatologies with Snowfall Data
Snowfall data needs to be added as a variable that can be accepted by `generate_climos`.  Furthermore, `convert_pr_var_units(...)` needs to be extended to handle `prsn` data.

nikola-rados:
Resolves #65 by adding `prsn` into supported variables.  Furthermore, extended `convert_pr_var_units` to `convert_flux_var_units` to handle `prsn`.  New test file added with snowfall data and some tests were updated.

### DIFF
--- a/dp/generate_climos.py
+++ b/dp/generate_climos.py
@@ -389,18 +389,20 @@ def convert_flux_var_units(input_file, climo_data):
     """If the file contains a 'pr' or 'prsn' variable, and if its units are per
        second, convert its units to per day.
     """
-    attributes = {}  # will contain updates, if any, to pr variable attributes
     flux_vars = [var for var in ['pr', 'prsn'] if var in input_file.dependent_varnames()]
 
     for flux_var in flux_vars:
+        attributes = {}  # will contain updates, if any, to flux variable attributes
         flux_variable = input_file.variables[flux_var]
         units = Unit.from_udunits_str(flux_variable.units)
+
         if units in [Unit('kg / m**2 / s'), Unit('mm / s')]:
             logger.info("Converting {} variable to units mm/day".format(flux_var))
             # Update units attribute
             attributes['units'] = (units * Unit('s / day')).to_udunits_str()
             # Multiply values by 86400 to convert from mm/s to mm/day
             seconds_per_day = 86400
+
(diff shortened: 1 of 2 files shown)


dp/generate_climos.py
```diff
@@ -383,44 +385,52 @@ def convert_longitude_range(climo_data):
...
+    attributes = {}  # will contain updates, if any, to pr variable attributes
+
+    vars = ['pr', 'prsn']
+    varname = ''
+    variable = None
+
+    for var in vars:
+        if var in input_file.dependent_varnames():
+            varname = var
+            variable = input_file.variables[var]
```
corviday: This does not handle the case where a file contains _both_ a `pr` variable and a `prsn` variable.

nikola-rados: Oh ok! I didn't know there would be a case where that was possible.  I'll make the changes!

corviday: It seems unlikely given the way prsn is generated with a separate script right now, to be sure. We do sometimes process files with multiple variables in them (hydrology's models often operate on several variables at once, and hydrology could make prsn data someday), so it's best not to taunt the fates.

nikola-rados:
@corviday let me know if there is anything else!

dp/generate_climos.py
```diff
@@ -383,44 +385,45 @@ def convert_longitude_range(climo_data):
...
+                var_only = cdo.select('name={}'.format(flux_var), input=climo_data)
                 # Multiply values by 86400 to convert from mm/s to mm/day
-                pr_only = cdo.mulc(str(seconds_per_day), input=pr_only)
+                var_only = cdo.mulc(str(seconds_per_day), input=var_only)
                 # Replace pr in all-variables file
-                climo_data = cdo.replace(input=[climo_data, pr_only])
+                climo_data = cdo.replace(input=[climo_data, var_only])
 
     # Update pr variable metadata as necessary to reflect changes madde
     with CFDataset(climo_data, mode='r+') as cf:
```
corviday: I think you're only updating attributes for one of the variables in cases where there are two. This `with` should be inside the `for`, and possibly `attributes` needs to be reset to empty after each variable.

nikola-rados: Ah you're right, my mistake there.

PR #76, opened 2019-06-10
Adds climatology period option to generate_climos

### LINKED ISSUE #74: Add a climatological periods argument to generate_climos
By default, the `generate_climos` script creates climatlogies for *all* of the periods available in the input file. This is great when starting from scratch. However there would be use cases when we're infilling climatologies (e.g. after a failure mid-script) where we want to generate *some* of the periods, but not all.

We should add a command line flag (like the one that is already sketched out [here](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/scripts/generate_climos#L47)) that allow the user to select only the periods that they want. All by default.

jameshiebert:
Resolves #74 The `-c --climo` option can be used multiple times, once for each climatological period that you wish to compute. Defaults to all. Options are included in the help and are gathered from the "standard" climatology names in ``nchelpers``.

This is pretty straightforward, but does anyone have time to briefly look over this.

### DIFF
--- a/scripts/generate_climos
+++ b/scripts/generate_climos
@@ -3,7 +3,7 @@ from argparse import ArgumentParser
 import logging
 import sys
 
-from nchelpers import CFDataset
+from nchelpers import CFDataset, standard_climo_periods
 from dp.argparse_helpers import strtobool, log_level_choices
 from dp.generate_climos import logger, create_climo_files
 
@@ -18,7 +18,8 @@ def main(args):
             except Exception as e:
                 logger.info('{}: {}'.format(e.__class__.__name__, e))
             else:
-                logger.info('climo_periods: {}'.format(input_file.climo_periods.keys()))
+                periods = input_file.climo_periods.keys() & args.climo
+                logger.info('climo_periods: {}'.format(periods))
                 for attr in 'project institution model emissions run'.split():
                     try:
                         logger.info('{}: {}'.format(attr, getattr(input_file.metadata, attr)))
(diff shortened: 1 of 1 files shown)


scripts/generate_climos
```diff
@@ -18,7 +18,8 @@ def main(args):
             except Exception as e:
                 logger.info('{}: {}'.format(e.__class__.__name__, e))
             else:
-                logger.info('climo_periods: {}'.format(input_file.climo_periods.keys()))
+                periods = input_file.climo_periods.keys() & args.climo
+                logger.info('climo_periods: {}'.format(periods))
```
rod-glover: Isn't `args.climo` a list? The `&` operator isn't defined when used in an expression with types `dict_keys` and `list` respectively. It doesn't like `list`, though it does like `dict_keys`.

scripts/generate_climos
```diff
@@ -18,7 +18,8 @@ def main(args):
             except Exception as e:
                 logger.info('{}: {}'.format(e.__class__.__name__, e))
             else:
-                logger.info('climo_periods: {}'.format(input_file.climo_periods.keys()))
+                periods = input_file.climo_periods.keys() & args.climo
```
rod-glover: `args.climo` is a list if the user specifies it (but not if not; you supply the default value of `standard_climo_periods().keys()`, which is of type `dict_keys`. The operator `&` isn't defined for operands of type `list`, though it does work as expected for `dict_keys`.

rod-glover:
A little problem with type (see comment). Otherwise LGTM.

scripts/generate_climos
jameshiebert: I'm pretty sure that it works as long as one argument is a `dict_keys`, no?

```
$ python3
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> {1: 1, 2: 2, 3: 3}.keys() & [1, 2]
{1, 2}
>>> [1, 2] & {1: 1, 2: 2, 3: 3}.keys()
{1, 2}
>>> [1, 2] & {1: 1, 2: 2, 3: 3}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for &: 'list' and 'dict'
```

rod-glover: Okay, that's weird. I got an error the first time I tried this, but now it shows you are right. Not a good day for me.

rod-glover: Ah, I forgot the function invocation parentheses. Definitely doesn't work with a signature of `function` & `list` :) :(

jameshiebert:
Cool, thanks guys!

PR #78, opened 2019-06-21
Add the ability to pipe YAML to update_metadata

jameshiebert:
In the Unix world, it's often useful to pipe input to a process on
STDIN especially when one is scripting things. This change adds the
ability for the update_metadata script to take input on STDIN if the
"-u -" option is used.


rod-glover:
Good idea. This looks fine to me. But I wonder if it would be more idiosyncratic to by default take input from STDIN and from a file only when the `-u` command is specified?

PR #81, opened 2019-08-22
Add aggregation of maximum and minimum climdex variables

corviday:
Previously (https://github.com/pacificclimate/climate-explorer-data-prep/pull/61), the `generate_climos` script was updated to correctly handle variables that count the number of times something happens in a specific time period, like `fdETCCDI` (frost days), which counts the number of days where the temperature goes below zero. The number of frost days in a year is the sum of the number of frost days in each month of that year. In order to calculate the average number of frost days per year over a 30-year period, we:
1. calculate the number per year by adding all the per-month values for that year together
2. calculate the long term average by taking the mean of all the yearly values

This PR expands that framework to also handle variables that measure the minimum or maximum of something, like `tnnETCCDI`, which measure the lowest temperature recorded each month. The yearly value for `tnnETCCDI` should be the lowest temperature recorded for that year. `rx1dayETCCDI` measures the maximum amount of precipitation that falls in a single day in each month; the value of `rx1dayETCCDI` for a year should the be maximum precipitation that falls in one day of that year.


dp/generate_climos.py
```diff
@@ -248,18 +255,19 @@ def climo_outputs(time_resolution, operation, data, aggregate=True,
...
+    elif var_type in ['count', 'maximum', 'minimum']:
         # cdo's ymon* climatology-creating operations assume means for
         # constructing subyear aggregate datasets (like making a monthly
         # dataset from a daily dataset).
-        # Count variables should be summed, not meaned, to aggregate within a
-        # year. Construct the aggregate datasets explicitly with a separate
-        # cdo command, then calculate climatologies - disallowing further
-        # aggregation - from them.
-        aggregations = generate_counted_aggregates(temporal_subset, input_file)
+        # For these variables, the value of a year is *not* the mean of the
```
nikola-rados: Great documentation as always!

Issue #71, opened 2019-05-03
prsn data should be in cm, not mm


corviday:
According to Trevor, precipitation as snow is normally given in centimeters, so having our prsn data in centimeters as well (instead of mm or kg/m2/day) will better communicate that we are dealing with snow.

PR #113, opened 2020-05-11
Convert prsn data from cm to mm

eyvorchuk:
The `generate_climos.py` and `generate_prsn.py` files have been modified to convert the prsn data from `mm` to `cm` (equivalently, `kg/m2` to `g/cm2`). The `test_generate_prsn.py` file has an additional assert statement to ensure the correct units.


dp/generate_prsn.py
```diff
@@ -166,8 +167,14 @@ def check_pr_units(pr_units):
...
+	Unit('mm / d'),
         Unit('kg / d / m**2'),
-        Unit('kg / m**2 / d')
+        Unit('kg / m**2 / d'),
+	Unit('g / cm**2 / s'),
+	Unit('cm / s'),
+	Unit('cm / d'),
+	Unit('g / d / cm**2'),
+	Unit('g / cm**2 / d')
     ]
```
nikola-rados, line 178: It may be helpful to order these in some manner (maybe grouping by `mm`, `cm`, `g`, `kg`) just to make it easier to follow.  You may want to double check with @corviday to ensure we aren't missing any cases here.

dp/generate_prsn.py
```diff
@@ -222,6 +229,11 @@ def process_to_prsn(variables, output_dataset, max_chunk_len):
             }
         means = np.mean([chunk_data['tasmin'], chunk_data['tasmax']], axis=0)
         prsn_data = np.where(means < freezing, chunk_data['pr'], 0)
+
+        prsn_units_from = ureg.parse_units('kg/m^2/s')
+        prsn_units_to = ureg.parse_units('g/cm^2/s')
+        prsn_data = prsn_data * Q_(1.0, prsn_units_from).to(prsn_units_to).magnitude	
+
         output_dataset.variables['prsn'][chunk] = prsn_data
```
nikola-rados, line 237: Have you considered skipping the extra variable assignment and assigning directly?

```python3
output_dataset.variables['prsn'][chunk] = prsn_data * Q_(1.0, prsn_units_from).to(prsn_units_to).magnitude
```

dp/generate_climos.py
```diff
@@ -442,11 +442,14 @@ def convert_flux_var_units(input_file, climo_data):
         flux_variable = input_file.variables[flux_var]
         units = Unit.from_udunits_str(flux_variable.units)
 
-        if units in [Unit('kg / m**2 / s'), Unit('mm / s')]:
-            logger.info("Converting {} variable to units mm/day".format(flux_var))
+        if units in [Unit('kg / m**2 / s'), Unit('mm / s'), Unit('g / cm**2 / s'), Unit('cm / s')]:
```
jameshiebert, line 445: An option that would make this more readable is to define each set of units as a variable to use in your conditional. For example:

```python3
precip_units = {Unit('kg / m**2 / s'), Unit('mm / s'), Unit('g / cm**2 / s'), Unit('cm / s')}
snow_units = {Unit('g / m**2 / s'), Unit('cm / s')}
if units in precip_units:
    ...
```

dp/generate_climos.py
```diff
@@ -442,11 +442,14 @@ def convert_flux_var_units(input_file, climo_data):
...
+            if units in [Unit('kg / m**2 / s'), Unit('mm / s')]:
```
jameshiebert, line 446: Do you need this conditional? It only changes one thing... the output unit in the log message. But you already *have* the string for the output unit string below: `(units * Unit('s / day')).to_udunits_str()`. Save that here, then use it both for your log message and for your attributes. Then the conditional can go away.

See ["If Statements are a Code Smell"](https://us.pycon.org/2020/schedule/presentation/87/).

dp/generate_prsn.py
```diff
@@ -222,6 +229,11 @@ def process_to_prsn(variables, output_dataset, max_chunk_len):
             }
         means = np.mean([chunk_data['tasmin'], chunk_data['tasmax']], axis=0)
         prsn_data = np.where(means < freezing, chunk_data['pr'], 0)
+
+        prsn_units_from = ureg.parse_units('kg/m^2/s')
+        prsn_units_to = ureg.parse_units('g/cm^2/s')
```
jameshiebert, line 234: Can you tell me more about using this as the target unit? You've changed both the mass (kg -> g) and the area (per square meter to per square centimeter). Is that definitely what we want? And is it definitely correct?

tests/test_generate_prsn.py
```diff
@@ -67,7 +67,7 @@ def test_create_prsn_netcdf_from_source(tiny_dataset, fake_dataset):
     assert np.shape(fake_dataset.variables['prsn']) == (11688, 4, 2)
     assert fake_dataset.variables['prsn'].standard_name == 'snowfall_flux'
     assert fake_dataset.variables['prsn'].long_name == 'Precipitation as Snow'
-
+    assert fake_dataset.variables['prsn'].units == 'g cm-2 s-1'
```
jameshiebert, line 70: There's not actually any check on the data values in this test, so it's hard to know whether it's doing the conversion correctly. Could you work out what the sum snowfall in this data *should* be an add that to the test?

dp/generate_prsn.py
eyvorchuk: Converting from kg -> g multiplies the data by 1000 and converting from /m^2 -> /cm^2 divides the data by 10000, giving the net effect of dividing the data by 10. Since converting from mm to cm also divides by 10, I think this will work. I can check with Lee when he's back.

tests/test_generate_prsn.py
eyvorchuk: I believe the dataset used for `test_process_to_prsn` returns a prsn array of all zeros since these tests pass.
```python
result = fake_dataset.variables['prsn'][:] 
result = np.where(result != 0)
for array in result:
    assert len(array) == 0
```
Is there possibly another dataset I could use to test the conversion?

jameshiebert: Oh right, you had mentioned that. Well that's not a great testing dataset, then is it?! :)

Maybe you could throw a few strategic low temperature values into the test data set... it might throw some other tests out of whack, but you could adjust as necessary.

eyvorchuk: Since we're only concerned about the conversion, would it be ok if I store the `pr` dataset in two variables, perform the conversion on one of them, and check to make sure that each value in the two datasets differs by a factor of 10? Or would you prefer I modify the data set and tests where needed?

rod-glover:
I have a question: Why convert units? P2A, at least, converts units on the frontend; that is a presentation issue, which is the responsibility of the UI, not of the data backend. Possibly the Marmot may need this, possibly lacking units conversions? (That is easily remedied, particularly given the utilities now in P2A.)

eyvorchuk:
> I have a question: Why convert units? P2A, at least, converts units on the frontend; that is a presentation issue, which is the responsibility of the UI, not of the data backend. Possibly the Marmot may need this, possibly lacking units conversions? (That can be remedied, particularly given the utilities now in P2A.)

I'm not sure. @corviday wrote [the original issue](https://github.com/pacificclimate/climate-explorer-data-prep/issues/71).

corviday:
I forgot I'd written that issue. At the time I wrote it, we did not yet have p2a, just PCEX, so it didn't occur to me that we might do unit conversion in the front end instead.

Issue #82, opened 2019-10-17
generate_climos should update the history metadata attribute


corviday:
I have receently been trying to figure out which file a suspicious climatology was made from; it would be nice if generate_climos updated the `history` attribute with the command used to run it, similar to the way `cdo` does.

corviday:
Or maybe some other, new, optional attribute.

rod-glover:
Off the top, I think updating `history` is probably best. That's what it's for, and I think it would be best to avoid introducing anything unnecessary and non-standard unless we have to. Not updating `history` was a failing of the original code that I never quite got around to fixing, or even creating an issue for. Now, voila.

jameshiebert:
If we can recommend some specific as to what entries to add to the `history` attribute, this would make a `good-first-issue` for incoming developers.

corviday:
I don't think we have a formal standard for this, but I write scripts to make a `cdo`-style update to `history` of the general form `datetime: tool-name arguments\n` at the top of the `history` attribute, retaining all older entries.

[Here's a recent one from the Frost Free Day script](https://github.com/pacificclimate/data-prep-actions/blob/51495d444d57e258f900d5aaf17b70b3a868e1e2/actions/frost_free_days/frost_free_days.py#L51-L52).

sum1lim:
I figured that `generate_climos` actually updates the history attribute.
For example, the history attribute of `tiny_daily_pr.nc` is:
```
		:history = "Thu Mar 21 14:49:01 2019: cdo sellonlatbox,216.5625,219.375,40.4636506825932,48.8352434707287 /storage/data/climate/downscale/BCCAQ2/CMIP5/CanESM2/pr_day_CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc /storage/home/nrados/tiny_daily_pr.nc\n",
			"Thu Sep  1 14:34:03 2016: ncrcat -O /storage/data/climate/downscale/BCCAQ2/CMIP5/timetmp/time_subset_pr_day_CanESM2_historical_r1i1p1_19500101-20051231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/timetmp/time_subset_pr_day_CanESM2_rcp26_r1i1p1_20060101-21001231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/CanESM2/pr_day_CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc\n",
[...]

Issue #11, opened 2017-09-26
Some software has issues with non-monotonic longitudes in netCDF file


corviday:
NetCDF files are sometimes generated with longitudes that go from 0 to 180 and then from -180 to 0, like this:

```bash
ncdump -v lon tasmax_aClim_CanESM2_historical_r3i1p1_19610101-19901231.nc 

netcdf tasmax_aClim_CanESM2_historical_r3i1p1_19610101-19901231 {
dimensions:
    ...

// global attributes:
   ...

data:

 lon = 0, 2.8125, 5.625, 8.4375, 11.25, 14.0625, 16.875, 19.6875, 22.5, 
    25.3125, 28.125, 30.9375, 33.75, 36.5625, 39.375, 42.1875, 45, 47.8125, 
    50.625, 53.4375, 56.25, 59.0625, 61.875, 64.6875, 67.5, 70.3125, 73.125, 
    75.9375, 78.75, 81.5625, 84.375, 87.1875, 90, 92.8125, 95.625, 98.4375, 
    101.25, 104.0625, 106.875, 109.6875, 112.5, 115.3125, 118.125, 120.9375, 
    123.75, 126.5625, 129.375, 132.1875, 135, 137.8125, 140.625, 143.4375, 
    146.25, 149.0625, 151.875, 154.6875, 157.5, 160.3125, 163.125, 165.9375, 
    168.75, 171.5625, 174.375, 177.1875, -180, -177.1875, -174.375, 
    -171.5625, -168.75, -165.9375, -163.125, -160.3125, -157.5, -154.6875, 
    -151.875, -149.0625, -146.25, -143.4375, -140.625, -137.8125, -135, 
    -132.1875, -129.375, -126.5625, -123.75, -120.9375, -118.125, -115.3125, 
    -112.5, -109.6875, -106.875, -104.0625, -101.25, -98.4375, -95.625, 
    -92.8125, -90, -87.1875, -84.375, -81.5625, -78.75, -75.9375, -73.125, 
    -70.3125, -67.5, -64.6875, -61.875, -59.0625, -56.25, -53.4375, -50.625, 
    -47.8125, -45, -42.1875, -39.375, -36.5625, -33.75, -30.9375, -28.125, 
    -25.3125, -22.5, -19.6875, -16.875, -14.0625, -11.25, -8.4375, -5.625, 
    -2.8125 ;
}
```

[...]

PR #117, opened 2020-05-25
NetCDF history attribute indicates when generate_climos started and e…

sum1lim:
…nded

As described in issue[ #82](https://github.com/pacificclimate/climate-explorer-data-prep/issues/82), `generate_climos` did not update history attribute for its own command(while it still updated `cdo` commands ran inside `generate_climos`). For example:
```
			"Fri May 22 15:21:55 2020: cdo -O -replace /tmp/cdoPyy5na14qm /tmp/cdoPy4pzh5rv7 /tmp/cdoPy6exys6hk\n",
			"Fri May 22 15:21:54 2020: cdo -O -timmean /tmp/cdoPysbe3hq0e /tmp/cdoPyy5na14qm\n",
			"Fri May 22 15:21:54 2020: cdo -O -seldate,1961-01-01,1990-12-31 tests/data/tiny_daily_pr.nc /tmp/cdoPysbe3hq0e\n",
			"Thu Mar 21 14:49:01 2019: cdo sellonlatbox,216.5625,219.375,40.4636506825932,48.8352434707287 /storage/data/climate/downscale/BCCAQ2/CMIP5/CanESM2/pr_day_CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc /storage/home/nrados/tiny_daily_pr.nc\n",
			"Thu Sep  1 14:34:03 2016: ncrcat -O /storage/data/climate/downscale/BCCAQ2/CMIP5/timetmp/time_subset_pr_day_CanESM2_historical_r1i1p1_19500101-20051231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/timetmp/time_subset_pr_day_CanESM2_rcp26_r1i1p1_20060101-21001231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/CanESM2/pr_day_CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc\n",
			"Thu Sep 01 14:33:15 2016: cdo -O seldate,1950-01-01T00:00,2005-12-31T23:59 /storage/data/climate/downscale/BCCAQ2/CMIP5/spacetmp/space_subset_pr_day_CanESM2_historical_r1i1p1_18500101-20051231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/timetmp/time_subset_pr_day_CanESM2_historical_r1i1p1_19500101-20051231.nc\n",
			"Thu Sep  1 14:32:42 2016: ncks -O -d lon,215.,310. -d lat,40.,85. /storage/data/climate/downscale/BCCAQ2/CMIP5/grouptmp/pr_day_CanESM2_historical_r1i1p1_18500101-20051231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/spacetmp/space_subset_pr_day_CanESM2_historical_r1i1p1_18500101-20051231.nc\n",
			"2011-04-13T23:04:41Z CMOR rewrote data to comply with CF standards and CMIP5 requirements.\n",
			"" ;
```
the last 3 `cdo` commands ran as a part of `generate_climos` command,  but it is almost impossible to know the information(time & arguments) of `generate_climos` was performed on. Therefore, additional lines that describe `generate_climos' command were desired to be present in the history attribute. This commit satisfies the desired functionality by generating the following result:

```
		:history = "Fri May 22 15:21:55 2020: end : generate_climos -otests/output tests/data/tiny_daily_pr.nc -p mean\n",
			"Fri May 22 15:21:55 2020: cdo -O -replace /tmp/cdoPyy5na14qm /tmp/cdoPy4pzh5rv7 /tmp/cdoPy6exys6hk\n",
			"Fri May 22 15:21:54 2020: cdo -O -timmean /tmp/cdoPysbe3hq0e /tmp/cdoPyy5na14qm\n",
			"Fri May 22 15:21:54 2020: cdo -O -seldate,1961-01-01,1990-12-31 tests/data/tiny_daily_pr.nc /tmp/cdoPysbe3hq0e\n",
			"Fri May 22 15:21:54 2020: start : generate_climos -otests/output tests/data/tiny_daily_pr.nc -p mean\n",
			"Thu Mar 21 14:49:01 2019: cdo sellonlatbox,216.5625,219.375,40.4636506825932,48.8352434707287 /storage/data/climate/downscale/BCCAQ2/CMIP5/CanESM2/pr_day_CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc /storage/home/nrados/tiny_daily_pr.nc\n",
			"Thu Sep  1 14:34:03 2016: ncrcat -O /storage/data/climate/downscale/BCCAQ2/CMIP5/timetmp/time_subset_pr_day_CanESM2_historical_r1i1p1_19500101-20051231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/timetmp/time_subset_pr_day_CanESM2_rcp26_r1i1p1_20060101-21001231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/CanESM2/pr_day_CanESM2_historical+rcp26_r1i1p1_19500101-21001231.nc\n",
			"Thu Sep 01 14:33:15 2016: cdo -O seldate,1950-01-01T00:00,2005-12-31T23:59 /storage/data/climate/downscale/BCCAQ2/CMIP5/spacetmp/space_subset_pr_day_CanESM2_historical_r1i1p1_18500101-20051231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/timetmp/time_subset_pr_day_CanESM2_historical_r1i1p1_19500101-20051231.nc\n",
			"Thu Sep  1 14:32:42 2016: ncks -O -d lon,215.,310. -d lat,40.,85. /storage/data/climate/downscale/BCCAQ2/CMIP5/grouptmp/pr_day_CanESM2_historical_r1i1p1_18500101-20051231.nc /storage/data/climate/downscale/BCCAQ2/CMIP5/spacetmp/space_subset_pr_day_CanESM2_historical_r1i1p1_18500101-20051231.nc\n",
			"2011-04-13T23:04:41Z CMOR rewrote data to comply with CF standards and CMIP5 requirements.\n",
			"" ;
```
Yet, the command line arguments shown in the history only include `-o; --outdir` `filepaths`. Adding more arguments would require passing `args` or more parameters when `create_climo_files` is called in `generate_climos` script. I would like to hear any suggestion if there are more necessary arguments to be shown in history attribute.

For the record, `black` command has not been applied to the changed files yet.

### DIFF
--- a/dp/generate_climos.py
+++ b/dp/generate_climos.py
@@ -7,6 +7,7 @@ import warnings
 from datetime import datetime
 import dateutil.parser
 import numpy as np
+import re
 
 from cdo import Cdo
 from netCDF4 import date2num
@@ -33,7 +34,7 @@ logger.setLevel(logging.DEBUG)  # For testing, overridden by -l when run as a sc
 cdo = Cdo()
 
 
-def create_climo_files(outdir, input_file, operation, t_start, t_end,
+def create_climo_files(outdir, filepath, input_file, operation, t_start, t_end,
                        convert_longitudes=True, split_vars=True, split_intervals=True,
                        output_resolutions={'yearly', 'seasonal', 'monthly'}):
     """Generate climatological files from an input file and a selected time range.
@@ -187,6 +188,15 @@ def create_climo_files(outdir, input_file, operation, t_start, t_end,
         'prsn': 'point'
     }
 
+    def generate_curr_time():
+        """This function returnscurrent time in the format of
+        cdo's history attribute.
+        """
+        return datetime.now().strftime("%a %B %d %H:%M:%S %Y")
+
+    # Record the starting time of generate_climos
+    t_start_generate_climos = generate_curr_time()
+
     try:
         var_types = {variable_types[variable] for variable in input_file.dependent_varnames()}
     except KeyError as e:
@@ -288,10 +298,13 @@ def create_climo_files(outdir, input_file, operation, t_start, t_end,
     # Convert units on any pr variable in each file
     climo_files = [convert_flux_var_units(input_file, climo_file) for climo_file in climo_files]
 
+    # Record the ending time of generate_climos
+    t_end_generate_climos = generate_curr_time()
+
     # Update metadata in climo files
     logger.debug('Updating climo metadata')
-    climo_files = [update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_file)
-                         for climo_file in climo_files]
+    climo_files = [update_metadata_and_time_var(input_file, outdir, filepath, t_start, t_end, operation, climo_file, t_start_generate_climos, t_end_generate_climos,
+                                                    split_vars, split_intervals) for climo_file in climo_files]
 
     # Split climo files by dependent variables if required
     if split_vars:
@@ -484,7 +497,8 @@ def split_on_variables(climo_file, var_names):
         return [climo_file]
 
 
-def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_filepath):
+def update_metadata_and_time_var(input_file, outdir, filepath, t_start, t_end, operation, climo_filepath, t_start_generate_climos, t_end_generate_climos, 
+                                    split_vars, split_intervals):
     """Updates an existing netCDF file to reflect the fact that it contains climatological means or standard deviations.
 
     Specifically:
@@ -515,6 +529,100 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
         # Update tracking_id attribute
         if hasattr(input_file, 'tracking_id'):
             cf.climo_tracking_id = input_file.tracking_id
+
+        def compare_times(t1, t2):
+            """This function takes two string inputs which are time representations
+            and returns boolean value. The purpose of this function is to find the
+            place where the start of generate_climos should be put in the history log.
+            It returns True if first input is later than the second input. Otherwise,
+            it returns False. If the second input is not in the format desired, the
+            function automatically returns True.
+            """
+            months = {
+                "Jan": 1,
+                "Feb": 2,
+                "Mar": 3,
+                "Apr": 4,
+                "May": 5,
+                "Jun": 6,
+                "Jul": 7,
+                "Aug": 8,
+                "Sep": 9,
+                "Oct": 10,
+                "Nov": 11,
+                "Dec": 12,
+            }
+            t1 = re.split(":|  | ", t1)
+            t2 = re.split(":|  | ", t2)
+
+            t1 = datetime(
+                int(t1[6]),
+                months[t1[1]],
+                int(t1[2]),
+                int(t1[3]),
+                int(t1[4]),
+                int(t1[5]),
+            )
+            try:
+                t2 = datetime(
+                    int(t2[6]),
+                    months[t2[1]],
+                    int(t2[2]),
+                    int(t2[3]),
+                    int(t2[4]),
+                    int(t2[5]),
+                )
+            except:
+                return True
+
+            return t1 > t2
+
+        def update_hisotry(history, start, end):
+            """This function takes the start time and end time history-attribute-lines
+            of generate_climos command. It returns a string of the entire history
+            attribute. The result indicates when the generate_climos started and ended.
+            For example:
+            :history = "Thu May 21 11:12:05: end: generate_climos ..." <--- end of genereate_climos
+            "Thu May 21 11:12:05 2020: cdo -O -replace ..."
+                "Thu May 21 11:12:04 2020: cdo -O -timmean ..." 
+                "Thu May 21 11:12:04 2020: cdo -O -seldate ..."
+                "Thu May 21 11:12:04: start: generate_climos ..." <--- start of genereate_climos
+                "Thu Mar 21 14:49:01 2019: cdo sellonlatbox ..."
+                "Thu Sep  1 14:34:03 2016: ncrcat ..."
+                "Thu Sep 01 14:33:15 2016: cdo -O seldate ..."
+                
+                ...
+            """
+            history_li = history.split("\n")
+            for idx, h in enumerate(history_li):
+                if compare_times(t_start_generate_climos, h[:24]):
+                    break
+                if idx == len(history_li)-1:
+                    idx += 1
+            history_li.insert(idx, start)
+
+            history_str = ""
+            for h in history_li:
+                history_str += h + "\n"
+
+            history_str = end + "\n" + history_str
+
+            return history_str
+
+
+        # Update history attribute
+        if hasattr(input_file, "history"):
+            command = ": generate_climos -o" + outdir + " " + filepath + " -p " + operation
+
+            start = (
+                t_start_generate_climos + ": start " + command
+            )
+            end = (
+                t_end_generate_climos + ": end " + command
+            )
+
+            cf.history = update_hisotry(cf.history, start, end)
+
 
         # Deduce the set of averaging intervals from the number of times in the file.
         # WARNING: This is fragile, and depends on the assumption that a climo output file contains only the following
--- a/scripts/generate_climos
+++ b/scripts/generate_climos
@@ -40,7 +40,7 @@ def main(args):
         else:
             for period in input_file.climo_periods.keys() & args.climo:
                 t_range = input_file.climo_periods[period]
-                create_climo_files(args.outdir, input_file, args.operation, *t_range,
+                create_climo_files(args.outdir, filepath, input_file, args.operation, *t_range,
                                    convert_longitudes=args.convert_longitudes, split_vars=args.split_vars,
                                    output_resolutions=args.resolutions)
 


dp/generate_climos.py
```diff
@@ -516,6 +530,100 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
...
 
+        def compare_times(t1, t2):
+            """This function takes two string inputs which are time representations
+            and returns boolean value. The purpose of this function is to find the
+            place where the start of generate_climos should be put in the history log.
+            It returns True if first input is later than the second input. Otherwise,
+            it returns False. If the second input is not in the format desired, the
+            function automatically returns True.
+            """
+            months = {
```
jameshiebert, line 541: How about [`calendar.month_abbr](https://docs.python.org/3/library/calendar.html#calendar.month_abbr) from the Python standard library?

dp/generate_climos.py
```diff
@@ -187,6 +189,15 @@ def create_climo_files(outdir, input_file, operation, t_start, t_end,
         'prsn': 'point'
     }
 
+    def generate_curr_time():
```
rod-glover, line 192: We could use a more specific name for this function. `generate_curr_time` is pretty generic. An alternative would be something like `curr_time_cdo_format`.

A side note: For functions (which are side-effect free, as opposed to procedures, which exist primarily for their side effects), we often prefer to omit verbs like "generate" or "make" from their names unless doing so would be confusing. Those verbs are often generic and amount to "compute", which is already implicit in the notion of a function.

dp/generate_climos.py
```diff
@@ -484,7 +498,8 @@ def split_on_variables(climo_file, var_names):
         return [climo_file]
 
 
-def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_filepath):
+def update_metadata_and_time_var(input_file, outdir, filepath, t_start, t_end, operation, climo_filepath, t_start_generate_climos, t_end_generate_climos,
```
rod-glover, line 501: I'm wondering if the approach here could be simplified: What if `generate_climos` just inserted something into the history before kicking off all the `cdo` operations, and finished up by inserting something afterwards? We would need to find a file in which to add the "start" history item; do you think that the initial selection of the temporal subset would be a simple place to do this?

corviday:
Overall, looks very good. I ran `generate_climos` on a bunch of files with weird history attributes, like date-last lines, histories with no dates at all, every kind of weirdness I can remember seeing in our data. It worked well on all of them.

However, if I run `generate_climos` on a file with no history attribute at all, the start and end lines do not appear. We hope any data files we process will have a history attribute, but they don't always, and it would be nice to support that case.

(If you would like to make a data file with no history attribute for testing, you can use the `update_metadata` script in this project to remove the history attribute from an existing file.)

Also, a very small formatting change; there's no space between `-o` and the output directory: 
`Mon May 25 11:49:10 2020: end : generate_climos -ooutput/ streamflow_peace_CanESM2_rcp85_r1i1p1.nc -p mean`

> I would like to hear any suggestion if there are more necessary arguments to be shown in history attribute.

It would be nice to have the `-g` argument if the user has specified one. The longitude normalization code [possibly generates files that some tools can't read](https://github.com/pacificclimate/climate-explorer-data-prep/issues/11), though investigating that issue has been a low priority. It might be useful to know which files are affected.

dp/generate_climos.py
```diff
@@ -516,6 +531,87 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
         if hasattr(input_file, 'tracking_id'):
             cf.climo_tracking_id = input_file.tracking_id
 
+        def compare_times(t1, t2):
```
rod-glover, line 534: We might be able to simplify this considerably extracting the string-to-datetime function embedded here, and use it just to the convert (date) strings extracted from history, and compare the result in-place, e.g., `start < to_datetime(s)`. That would make the logic clearer and DRY up this code, which contains code for the date conversion repeated twice, once for each argument.

And we could also not convert the start/end time forward into a string, then back into a datetime for comparison; We could just pass the start/end time as a `datetime`, compare it without need for conversion, and convert it just once when it is written to the history as a string. It's almost always better to keep values in their natural computational representation (in this case `datetime`) internally, and only convert to strings for I/O.

Tell me if I am missing something here.

dp/generate_climos.py
```diff
@@ -516,6 +531,87 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
...
+                    int(t2[3]),
+                    int(t2[4]),
+                    int(t2[5]),
+                )
+            except:
+                return True
+
+            return t1 > t2
+
+        def update_hisotry(history, start, end):
```
corviday, line 568: Typo in function name.

dp/generate_climos.py
```diff
@@ -33,7 +35,7 @@
 cdo = Cdo()
 
 
-def create_climo_files(outdir, input_file, operation, t_start, t_end,
+def create_climo_files(outdir, filepath, input_file, operation, t_start, t_end,
```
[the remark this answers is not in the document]
rod-glover: And can you also explain to me what `filepath` is for?

dp/generate_climos.py
```diff
@@ -583,6 +608,79 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
...
+            command += " -r " + str(output_resolutions).split("\'")[1]
+
+        if position:
+            hist_line = (time_cdo_format + ": start" + command)
+            hist_list = history.split("\n")
+            hist_list.insert(position, hist_line)
+
+            hist_str = ""
+            for h in hist_list:
+                hist_str += h + "\n"
```
rod-glover, line 675: There's an easier way to do this: the Python `join` function. It's for exactly this operation.

dp/generate_climos.py
```diff
@@ -583,6 +608,79 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
...
+            command += " -g " + str(convert_longitudes)
+
+        if not split_vars:
+            command += " -v " + str(split_vars)
+
+        if not split_intervals:
+            command += + " -i " + str(split_intervals)
+
+        if output_resolutions != set(["yearly", "seasonal", "monthly"]):
+            command += " -r " + str(output_resolutions).split("\'")[1]
```
rod-glover, line 666: I wonder if it would be better to use `sys.argv` to retrieve the command line arguments as entered rather than reconstructing them like this. It would certainly be less work. It would echo the actual original command line, rather than the defaults supplied. What are your thoughts?

dp/generate_climos.py
```diff
@@ -583,6 +608,79 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
...
 
+def update_generate_climos_history(input_filepath, outdir, operation, convert_longitudes, 
+                                    split_vars, split_intervals, output_resolutions, netCDF_file, time_cdo_format, position=0):
+    """This function takes the start time and end time history-attribute-lines
+    of generate_climos command. It returns a string of the entire history
+    attribute. The result indicates when the generate_climos started and ended.
+
+    :param netCDF_file: (CFDataset) netCDF file to be updated
+    :param time_cdo_format: (str) the point of time that generate_climos history(start or end) was updated
+    :param position: (int) the position that the new history attribute line will be inserted
```
rod-glover, line 619: That's clever. Maybe explain this slightly more, to the effect that it indexes into the lines of the history attribute.

dp/generate_climos.py
```diff
@@ -583,6 +608,79 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
...
+        if not split_vars:
+            command += " -v " + str(split_vars)
+
+        if not split_intervals:
+            command += + " -i " + str(split_intervals)
+
+        if output_resolutions != set(["yearly", "seasonal", "monthly"]):
+            command += " -r " + str(output_resolutions).split("\'")[1]
+
+        if position:
```
rod-glover, line 668: Testing for truthiness when you really mean (in this case) `> 0` can lead to problems: It's ambiguous, and, for example, it would let a negative number through as well. In general it is best to be explicit about your conditions: easier to understand, fewer bugs later. This case is pretty minor and unlikely to lead to error, but we have had many cases when an explicit test (e.g., `x is None` instead of `x`) would have saved all kinds of later trouble.

dp/generate_climos.py
```diff
@@ -583,6 +608,79 @@ def update_metadata_and_time_var(input_file, t_start, t_end, operation, climo_fi
...
+            hist_list = history.split("\n")
+            hist_list.insert(position, hist_line)
+
+            hist_str = ""
+            for h in hist_list:
+                hist_str += h + "\n"
+        
+        else:
+            hist_line = (time_cdo_format + ": end" + command)
+            hist_str = hist_line  + "\n" + history
```
rod-glover, line 679: Doesn't the `position != 0` case (`then` branch of `if`) handle this too?

dp/generate_climos.py
jameshiebert: :+1: for using `sys.argv`!

dp/generate_climos.py
```diff
@@ -736,8 +736,13 @@ def update_generate_climos_history(netCDF_file, time_cdo_format, position=0):
...
         command = " ".join(sys.argv)
-
-        hist_line = time_cdo_format + ": start" + command
+        
+        # update_generate_climos_history for the start has to be called after the first or latter cdo command
+        # update_generate_climos_history for the end has to be called after the last cdo command
+        if position != 0:
+            hist_line = time_cdo_format + ": start" + command
+        else:
+            hist_line = time_cdo_format + ": end" + command
```
rod-glover, line 745: Interesting. For the present usage, the inference of start/end from `position` is correct, in that it works. But it's not *necessarily* so; one can imagine (in a vague way at least) that there could be other desired use cases. In that case it would be worth adding an explicit boolean begin/end parameter to this method. For now, I suggest you just document this inference explicitly, in the docstring at the head of this function. Your example implies this, but it would be good to state it explicitly.

tests/test_create_climo_files.py
```diff
@@ -165,6 +169,21 @@ def test_climo_metadata(outdir, tiny_dataset, operation, t_start, t_end, split_v
...
 
+            # Tests for history metadata updates
+            assert cf.history
+            hist_lines = cf.history.split("\n")
+            end_items = history_items(hist_lines[0])
+            assert end_items[-1].split(" ")[0] == "generate_climos" and end_items[-2] == "end"
+            found_start = False
+            for idx, h in enumerate(hist_lines):
+                h_items = history_items(h)
+                if h_items[-2] == "start" and h_items[-1].split(" ")[0] == "generate_climos":
```
rod-glover, line 180: Wow, that was a lot of fun :stuck_out_tongue:

PR #121, opened 2020-06-04
Issue/119

eyvorchuk:
This PR adds the ability to output the generate_prsn dry run info to a text file rather than a logger. This is intended to be used for the wps generate_prsn process in thunderbird.

### DIFF
--- a/dp/generate_prsn.py
+++ b/dp/generate_prsn.py
@@ -19,9 +19,14 @@ Q_ = ureg.Quantity
 logger = logging.getLogger(__name__)
 
 
-def dry_run(filepaths, output_to_file=False, workdir=None):
-    '''Perform metadata checks on the input files'''
-    if output_to_file: # Used for wps process in thunderbird
+def dry_run(filepaths, outpath=None):
+    '''Perform metadata checks on the input files
+
+       Parameters:
+            filepaths (dict): Dictionary containing the three filepaths
+            outpath (str): Optional path to text file that will contain metadata info
+    '''
+    if outpath: # Used for wps process in thunderbird
         output_items = []
         outputer = output_items.append
     else:
@@ -42,13 +47,10 @@ def dry_run(filepaths, output_to_file=False, workdir=None):
                 outputer('{}: {}: {}'.format(attr, e.__class__.__name__, e))
         outputer('dependent_varnames: {}'.format(dataset.dependent_varnames()))
 
-    if output_to_file:
-        filename = os.path.join(workdir, 'dry.txt')
-        with open(filename, 'w') as f:
+    if outpath:
+        with open(outpath, 'w') as f:
             for line in output_items:
                 f.write('{}\n'.format(line))
-
-        return filename
 
 def unique_shape(arrays):
     '''Ensure each array in dict is the same shape'''
--- a/dp/generate_prsn.py
+++ b/dp/generate_prsn.py
@@ -22,45 +22,33 @@ logger = logging.getLogger(__name__)
 def dry_run(filepaths, output_to_file=False, workdir=None):
     '''Perform metadata checks on the input files'''
     if output_to_file: # Used for wps process in thunderbird
-        output = ['Dry Run']
-        for filepath in filepaths.values():
-            output.append('File: {}'.format(filepath))
+        output_items = []
+        outputer = output_items.append
+    else:
+        outputer = logger.info
+
+    outputer('Dry Run')
+    for filepath in filepaths.values():
+        outputer('File: {}'.format(filepath))
+        try:
+            dataset = CFDataset(filepath)
+        except Exception as e:
+            outputer('{}: {}'.format(e.__class__.__name__, e))
+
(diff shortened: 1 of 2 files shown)


dp/generate_prsn.py
```diff
@@ -19,23 +19,47 @@
...
-                logger.info('{}: {}: {}'.format(attr, e.__class__.__name__, e))
-        logger.info('dependent_varnames: {}'.format(dataset.dependent_varnames()))
+                logger.exception('{}: {}'.format(e.__class__.__name__, e))
+
+            for attr in 'project model institute experiment ensemble_member'.split():
+                try:
+                    logger.info('{}: {}'.format(attr, getattr(dataset.metadata, attr)))
+                except Exception as e:
+                    logger.info('{}: {}: {}'.format(attr, e.__class__.__name__, e))
+            logger.info('dependent_varnames: {}'.format(dataset.dependent_varnames()))
```
nikola-rados, line 62: While this works, I think we need to find a way to save on redundant code.  Both sides of the `if/else` run the exact same process with the exception of the way the logs are output.

What we might want to try in this case is defining an `outputer` that we can use in the same way for both cases. Here is an example of what I mean:
```python3
nrados@pcic-2024:~$ python3
Python 3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> import sys
>>> 
>>> logger = logging.getLogger()
>>> logger.setLevel(logging.DEBUG)
>>> 
>>> handler = logging.StreamHandler(sys.stdout)
>>> handler.setLevel(logging.DEBUG)
>>> formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
>>> handler.setFormatter(formatter)
>>> logger.addHandler(handler)
>>> 
>>> output_items = []
>>> 
>>> outputer = logger.info
>>> outputer('Foo')
2020-06-04 13:40:22,993 - root - INFO - Foo
>>> outputer = output_items.append
>>> outputer('Bar')
>>> output_items
['Bar']
```
You can see in both instances we passed a string into the method the same way but it produced a different result. With this `outputer` we would only have one set of dry run code.

dp/generate_prsn.py
```diff
@@ -19,24 +19,36 @@
...
 
-def dry_run(filepaths):
+def dry_run(filepaths, output_to_file=False, workdir=None):
     '''Perform metadata checks on the input files'''
-    logger.info('Dry Run')
+    if output_to_file: # Used for wps process in thunderbird
+        output_items = []
+        outputer = output_items.append
+    else:
+        outputer = logger.info
```
nikola-rados, line 28: Nice this is great!

dp/generate_prsn.py
```diff
@@ -19,24 +19,36 @@
 logger = logging.getLogger(__name__)
 
 
-def dry_run(filepaths):
+def dry_run(filepaths, output_to_file=False, workdir=None):
     '''Perform metadata checks on the input files'''
```
nikola-rados, line 23: You may want to consider adding a docstring here since the method has some new options others may not understand.

dp/generate_prsn.py
```diff
@@ -19,24 +19,36 @@
...
-                logger.info('{}: {}'.format(attr, getattr(dataset.metadata, attr)))
+                outputer('{}: {}'.format(attr, getattr(dataset.metadata, attr)))
             except Exception as e:
-                logger.info('{}: {}: {}'.format(attr, e.__class__.__name__, e))
-        logger.info('dependent_varnames: {}'.format(dataset.dependent_varnames()))
+                outputer('{}: {}: {}'.format(attr, e.__class__.__name__, e))
+        outputer('dependent_varnames: {}'.format(dataset.dependent_varnames()))
+
+    if output_to_file:
+        filename = os.path.join(workdir, 'dry.txt')
```
nikola-rados, line 46: We should have the file name configurable as a parameter. Instead of having a `workdir` and a `file` we may as well just define an `outpath`.  That way the path and filename is configurable for the user.

dp/generate_prsn.py
```diff
@@ -19,24 +19,36 @@
...
+        with open(filename, 'w') as f:
+            for line in output_items:
+                f.write('{}\n'.format(line))
 
+        return filename
```
nikola-rados, line 51: I'm not sure it is a good idea for us to return something from this method in some cases but not others.  I imagine you are using the `filename` in `thunderbird` to give as a `response`.  If that is the case I think adding the path as a param will solve the issue since you will already know the output location.

scripts/generate_prsn
```diff
@@ -23,14 +24,17 @@ def main(args):
         'tasmax': args.tasmax
     }
     if args.dry_run:
-        dry_run(filepaths)
+        outpath = os.path.join(args.outdir, args.dry_output_file) if args.dry_output_file else None
+        dry_run(filepaths, outpath)
```
nikola-rados, line 28: Looks good!

Issue #96, opened 2020-02-28
Remove use of subproces.call()


nikola-rados:
In `tests/test_decompose_flow_vectors.py` `subprocess.call()` is used to run other scripts.  This is causing issues in the jenkins pipeline.  In particular jenkins cannot find the `netCDF4` module once it opens a subprocess.  If we could find another way to test the scripts we can introduce them into the pipeline. Until then they will be excluded.

jameshiebert:
Do you have a recommended course of action? [The docs](https://docs.python.org/3.8/library/subprocess.html#older-high-level-api) for `subprocess.call()` mention that it's an old API and [`subprocess.run()`](https://docs.python.org/3.8/library/subprocess.html#subprocess.run) should be used.

Generally, using the `shell=True` argument preserves the caller's environment (which presumably would allow the callee to find the `netCDF4` module if it's available to the caller).

nikola-rados:
I don't have a recommended course of action but your suggestion of `shell=True` may do it. I think that is probably the best place to start.

corviday:
Alternately, we could consider moving `decompose_flow_vectors.py` to a different repository. It was written to generate versions of Hydrology's flow vector files that could be rendered by ncWMS as maps with arrows pointing from grid cell to grid cell where water flows. It hasn't become part of our primary data workflow, and might fit better in data-prep-actions with the rest of the standalone data processing scripts.

nikola-rados:
That's an interesting idea, I was not aware of the scripts origins.  If you are suggesting a more to data-prep-actions does that mean the script really only needs to be documented and not tested?

corviday:
`decompose_flow_vectors` predates Rod's excellent idea to establish data-prep-actions. Accordingly, it is more polished than usual for that repository, with tests and installation guides. It's limited usage, like the data prep actions scripts, but polished and tested, like the climate explorer data prep scripts. Could probably reasonably go in either location.

PR #122, opened 2020-06-11
Issue/118 Refactor `decompose_flow_vectors` for further WPS processes

### LINKED ISSUE #118: Refactor `decompose_flow_vectors`
The [`decompose_flow_vectors`](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/scripts/decompose_flow_vectors) script has no function encapsulation with all of the code being at the top level and executed at import time.

Better practice would be to wrap the functionality in a `main()` function, so that it can be imported into other programs and run programmatically (e.g. in a WPS process).

Follow the example in [`generate_prsn`](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/scripts/generate_prsn) to do this.

sum1lim:
This PR resolves #118 

For a better utility of the program, main function has been created and wraps the functionality of `decompose_flow_vectors`. The script and main function will be imported into other programs, thus the `scripts` package should be included in `ce_dataprep`. In ordered to do that `__init__.py` has been created.

### DIFF
--- a/dp/decompose_flow_vectors.py
+++ b/dp/decompose_flow_vectors.py
@@ -1,0 +1,68 @@
+#!python
+import argparse
+from netCDF4 import Dataset
+import numpy as np
+import time
+import sys
+import types
+
+def decompose_flow_vectors(source, dest, variable, input_filepath, output_filepath):
+
+    dest.createDimension("lat", source.dimensions["lat"].size)
+    dest.createVariable("lat", "f8", ("lat"))
+    dest.variables["lat"].setncatts(source.variables["lat"].__dict__)
+    dest.variables["lat"][:] = source.variables["lat"][:]
+
+    dest.createDimension("lon", source.dimensions["lon"].size)
+    dest.createVariable("lon", "f8", ("lon"))
+    dest.variables["lon"].setncatts(source.variables["lon"].__dict__)
+    dest.variables["lon"][:] = source.variables["lon"][:]
+
+    #create the vector variables
+    eastvec = "eastward_{}".format(variable)
+    dest.createVariable(eastvec, "f8", ("lat", "lon"))
+    dest.variables[eastvec].units = "1"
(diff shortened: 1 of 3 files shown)


scripts/decompose_flow_vectors
```diff
@@ -1,108 +1,61 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Process an indexed flow direction netCDF into a vectored netCDF suitable for ncWMS display"
+    )
+    parser.add_argument("source_file", metavar="infile", help="source netCDF file")
+    parser.add_argument("dest_file", metavar="outfile", help="destination netCDF file")
+    parser.add_argument(
+        "variable", metavar="variable", help="netCDF variable describing flow direction"
+    )
+
```
nikola-rados, line 59: We probably want to add a flag for log level.

setup.py
```diff
@@ -31,6 +31,7 @@
         scripts/generate_prsn
         scripts/split_merged_climos
         scripts/update_metadata
+	scripts/decompose_flow_vectors
```
nikola-rados, line 34: Just need two more spaces for this item to line up with the others in the list.

dp/decompose_flow_vectors.py
```diff
@@ -0,0 +1,135 @@
+#!python
```
nikola-rados, line 1: This entire file looks great!

corviday:
This script was [excluded from CI testing](https://github.com/pacificclimate/climate-explorer-data-prep/issues/96) because the way it was written made it difficult for the test system to call. That issue should be fixable now, but I'm not sure if we'd rather make it part of this PR, or do it separately.

rod-glover:
Thanks, @corviday for providing the ground-truth check.

sum1lim:
@corviday It is really interesting to see the actual output illustrated on a real map and glad to see the outputs are correct. Thank you! I guess this PR is ready to be merged, so I will do it right away. Thanks to all who made comments!

Issue #126, opened 2020-06-22
Refactor input checking procedures in decompose_flow_vectors script


sum1lim:
Currently, `decompose_flow_vectors script`'s [main(args)](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/scripts/decompose_flow_vectors#L18) contains input(source and variable) checking procedures. This is somewhat not ideal when the program is called in another program such as thunderbird. Passing `args` as a parameter makes it hard to call the `main()` function when the program that imports `decompose_flow_vectors` does not run using terminal input. Therefore, it is necessary to copy the input checking procedure from the script to the newly implemented program, which is redundant to have duplicate codes. To avoid this, refactor the procedure into `source_check()` and `variable_check()` functions.

PR #127, opened 2020-06-22
Issue/126 Refactor input validity checks in decompose_flow_vectors

sum1lim:
This PR closes issue #126 

For the convenience in importing `decompose_flow_vectors` in other programs such as WPS processes, the input validity checking procedure in main() has been refactored into `source_check` and `variable_check` functions. The advantage of the functions is that they do not need terminal input for arguments, thus they can be called in any programs that do not require terminal inputs.

The functions throw an user-defined exception if any error conditions are met, and main() `SystemExit`s the program when it catches an exception. There are two main user-defined exceptions(InvalidSourceError and InvalidVariableError) and 5 other exceptions that inherit the main exceptions. The reason for having many specific exceptions is to manage versatile error conditions when the functions are called in other programs

### DIFF
--- a/dp/decompose_flow_vectors.py
+++ b/dp/decompose_flow_vectors.py
@@ -23,14 +23,19 @@ def check_and_raise_exception(condition, source, err_msg, exc):
         raise exc(err_msg)
 
 
+def dimensions_validity(dimensions):
+    return not "lon" in dimensions or not "lat" in dimensions
+
+
 def source_check(source):
     source_file_path = source.filepath()
 
-    condition = not "lon" in source.dimensions or not "lat" in source.dimensions
     err_msg = "{} does not have latitude and longitude dimensions".format(
         source_file_path
     )
-    check_and_raise_exception(condition, source, err_msg, AttributeError)
+    check_and_raise_exception(
+        dimensions_validity(source.dimensions), source, err_msg, AttributeError
+    )
 
     valid_variables = []
     for v in source.variables:
@@ -57,11 +62,10 @@ def variable_check(source, variable):
 
     flow_variable = source.variables[variable]
 
-    condition = (
-        not "lon" in flow_variable.dimensions or not "lat" in flow_variable.dimensions
-    )
     err_msg = "Variable {} is not associated with a grid".format(variable)
-    check_and_raise_exception(condition, source, err_msg, AttributeError)
+    check_and_raise_exception(
+        dimensions_validity(flow_variable.dimensions), source, err_msg, AttributeError
+    )
 
     condition = np.ma.max(flow_variable[:]) > 9 or np.ma.min(flow_variable[:]) < 1
     err_msg = "Variable {} is not a valid flow routing".format(variable)
--- a/dp/decompose_flow_vectors.py
+++ b/dp/decompose_flow_vectors.py
@@ -16,20 +16,21 @@ logger.addHandler(handler)
 logger.setLevel(logging.DEBUG)  # For testing, overridden by -l when run as a script
 
 
-def log_and_raise_exception(source, err_msg, exc):
-    logger.critical(err_msg)
-    source.close()
-    raise exc(err_msg)
+def check_and_raise_exception(condition, source, err_msg, exc):
+    if condition:
+        logger.critical(err_msg)
+        source.close()
+        raise exc(err_msg)
 
 
 def source_check(source):
     source_file_path = source.filepath()
 
-    if not "lon" in source.dimensions or not "lat" in source.dimensions:
-        err_msg = "{} does not have latitude and longitude dimensions".format(
-            source_file_path
-        )
-        log_and_raise_exception(source, err_msg, AttributeError)
+    condition = not "lon" in source.dimensions or not "lat" in source.dimensions
+    err_msg = "{} does not have latitude and longitude dimensions".format(
+        source_file_path
+    )
+    check_and_raise_exception(condition, source, err_msg, AttributeError)
 
     valid_variables = []
     for v in source.variables:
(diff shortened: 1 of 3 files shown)


scripts/decompose_flow_vectors
```diff
@@ -14,38 +14,102 @@ import sys
...
+
+    try:
+        source_check(source)
+    except InvalidSourceError:
+        sys.exit(1)
+
+    try:
+        variable_check(source, args.variable)
+    except InvalidVariableError:
+        sys.exit(2)
```
jameshiebert, line 112: One one hand, I can see the merit to being as specific as possible (via the exit code) as to why the program exited. However, at this point (correct me if I'm wrong), I don't think that we're automating the use of this script from bash and there might not be any use for these specific exit code.

Which then raises the question as to whether there's a purpose for creating a class hierarchy above for the Exception types. Ever since attending this [provocatively named, and humorously informative PyCon talk](https://www.youtube.com/watch?v=o9pEzgHorH0) "Stop Writing Classes", I have been hyper-skeptical of custom exception types. Especially since there are a large number of built-in Python `Exception`s and that most `Exception` children allow instantiation with an informative message.

So I'd advise that you probably don't need almost 30 lines of custom Exception hierarchy and can probably do raising one of the [builtin Exceptions](https://docs.python.org/3/library/exceptions.html) (`TypeError`, `ValueError`, `RuntimeError`, ...) with an informative message.

sum1lim: This strategy has been selected based on my plan for applying the functions in [wps_decompose_flow_vectors](https://github.com/pacificclimate/thunderbird/blob/issue/12/thunderbird/processes/wps_decompose_flow_vectors.py#L123). I wanted the log message to appear on WPS environment as well, therefore the distinctions between error conditions were needed. However, instead of creating two main user-defined exceptions, it would be wiser to make the lower level exceptions to inherit built-in exceptions like you suggested.

jameshiebert: Yeah, having the log message appear in the WPS environment is absolutely an excellent use case. But I think that's what the `message` attribute of an `Exception` is for.

scripts/decompose_flow_vectors
```diff
@@ -14,38 +14,102 @@ import sys
...
 
     if np.ma.max(flow_variable[:]) > 9 or np.ma.min(flow_variable[:]) < 1:
-        logger.debug("Variable {} is not a valid flow routing".format(args.variable))
+        logger.critical("Variable {} is not a valid flow routing".format(variable))
         source.close()
-        sys.exit()
+        raise NotAVicModelError
+
+
+def main(args):
```
nikola-rados, line 100: Two things here:
1. We probably want to change the name of the main method to something more descriptive since we will be using it in another repo.
2. Knowing we will be using this in `thunderbird` we probably want to decompose the `args` param into individual components.

scripts/decompose_flow_vectors
nikola-rados: I would agree with James' comments here.

scripts/decompose_flow_vectors
sum1lim: I think we could maintain the `main()` function as it is now. The functions that `wps_decompose_flow_vectors` requires are `source_check`, `variable_check` and `dp.decompose_flow_vectors`, but `main()` won't be included. This `main()` function is specifically for `ce-dataprep` as far as I know. Please correct me if I am wrong.

nikola-rados: Ah ok sorry, I misread and thought this was in `dp/` and not `scripts/`.  That being said, if we want this to be importable to `thunderbird` these methods should be in `dp/decompose_vector_flows`.

rod-glover: I agree with @nikola-rados . The `scripts` code should be minimal, just arranging arguments and calling the function(s) in `dp/`.

dp/decompose_flow_vectors.py
```diff
@@ -16,6 +16,63 @@
...
+
+
+def source_check(source):
+    source_file_path = source.filepath()
+
+    condition = not dimensions_validity(source)
+    err_msg = "{} does not have latitude and longitude dimensions".format(
+        source_file_path
+    )
+    check_and_raise_exception(condition, source, err_msg, AttributeError)
```
nikola-rados, line 41: While I appreciate the cleverness of what you have done here, I _do_ think an `if` statement is more understandable. The way it is written right now we are _always_ creating an error message and _then_ checking if there is an error. I can imagine this being challenging to follow in the future.

scripts/decompose_flow_vectors
```diff
@@ -11,41 +11,28 @@ import logging
 from netCDF4 import Dataset
 import numpy as np
 import sys
-from dp.decompose_flow_vectors import logger, decompose_flow_vectors
+from dp.decompose_flow_vectors import (
+    logger,
+    decompose_flow_vectors,
+    source_check,
+    variable_check,
+)
```
nikola-rados, line 19: This sure cleans things up in here!

dp/decompose_flow_vectors.py
sum1lim: I had to do this because codeclimate did not like similar blocks repeating in a file. May I ignore what codeclimate suggests?

nikola-rados: > May I ignore what codeclimate suggests?

_Probably_ not. What specifically was it complaining about? Just the fact that we had two similar `if` statements that would build an `err_msg` then handle the exception?

sum1lim: It was because of these if statements:
[R28-R32](https://github.com/pacificclimate/climate-explorer-data-prep/pull/127/commits/93a5200a12b22c6cf0bf1f63a842ec8c7e22a58f#diff-1e7f88c622597ae3ca67157373b72288R28-R32)
[R59-R61](https://github.com/pacificclimate/climate-explorer-data-prep/pull/127/commits/93a5200a12b22c6cf0bf1f63a842ec8c7e22a58f#diff-1e7f88c622597ae3ca67157373b72288R59-R61)

nikola-rados: Ok, and what exactly was the complaint from codeclimate?

nikola-rados: Ok, I can see what it's complaining about.  That is a little bit annoying haha,  I think we _could_ ignore the error. We may be able to combine [these lines](https://github.com/pacificclimate/climate-explorer-data-prep/pull/127/commits/93a5200a12b22c6cf0bf1f63a842ec8c7e22a58f#diff-1e7f88c622597ae3ca67157373b72288R59-R65) into an `if/elif` statement, but I'm not sure if that would be enough. Maybe give that a try?

PR #133, opened 2020-07-08
Fixed bug #132. split_intervals passed into create_climo_files

### LINKED ISSUE #132: Input argument split_intervals should be passed into create_climo_files function call in generate_climos script
In `generate_climos` script, there exist an input option `split_intervals`, but it is never used elsewhere in the code. It is supposed to be passed into [create_climo_files](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/scripts/generate_climos#L48) as the function [expects](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/dp/generate_climos.py#L51) the argument to be passed.

sum1lim:
This PR closes #132 

As mentioned in the issue, the input option `split_intervals` exists but never passed into `create_climo_files`. Therefore, [split_variable was always set to default value "True"](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/dp/generate_climos.py#L51), which makes the same output files no matter what value was given to the option:
```
(venv) slim@devel2:~/Desktop/climate-explorer-data-prep$ generate_climos -o output_without_split_intervals ./tests/data/tiny_daily_pr.nc -p mean

...

(venv) slim@devel2:~/Desktop/climate-explorer-data-prep$ ls output_without_split_intervals
pr_aClimMean_CanESM2_historical_r1i1p1_19610101-19901231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_19710101-20001231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_19810101-20101231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_20100101-20391231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_20400101-20691231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_20700101-20991231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_19610101-19901231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_19710101-20001231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_19810101-20101231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_20100101-20391231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_20400101-20691231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_20700101-20991231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_19610101-19901231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_19710101-20001231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_19810101-20101231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_20100101-20391231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_20400101-20691231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_20700101-20991231.nc
```
```
(venv) slim@devel2:~/Desktop/climate-explorer-data-prep$ generate_climos -o output_no_argument_passing -i False ./tests/data/tiny_daily_pr.nc -p mean

...

(venv) slim@devel2:~/Desktop/climate-explorer-data-prep$ ls output_no_argument_passing
pr_aClimMean_CanESM2_historical_r1i1p1_19610101-19901231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_19710101-20001231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_19810101-20101231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_20100101-20391231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_20400101-20691231.nc
pr_aClimMean_CanESM2_historical_r1i1p1_20700101-20991231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_19610101-19901231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_19710101-20001231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_19810101-20101231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_20100101-20391231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_20400101-20691231.nc
pr_mClimMean_CanESM2_historical_r1i1p1_20700101-20991231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_19610101-19901231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_19710101-20001231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_19810101-20101231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_20100101-20391231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_20400101-20691231.nc
pr_sClimMean_CanESM2_historical_r1i1p1_20700101-20991231.nc
```

After modifying the code to pass the argument into `create_climo_files`, these are the output files I am getting:
```
(venv) slim@devel2:~/Desktop/climate-explorer-data-prep$ generate_climos -o output_argument_passing -i False ./tests/data/tiny_daily_pr.nc -p mean

...

(venv) slim@devel2:~/Desktop/climate-explorer-data-prep$ ls output_argument_passing
pr_msaClimMean_CanESM2_historical_r1i1p1_19610101-19901231.nc
pr_msaClimMean_CanESM2_historical_r1i1p1_19710101-20001231.nc
pr_msaClimMean_CanESM2_historical_r1i1p1_19810101-20101231.nc
pr_msaClimMean_CanESM2_historical_r1i1p1_20100101-20391231.nc
pr_msaClimMean_CanESM2_historical_r1i1p1_20400101-20691231.nc
pr_msaClimMean_CanESM2_historical_r1i1p1_20700101-20991231.nc
```


sum1lim:
This PR required a lot more modifications than expected. Therefore, I would like to ask @nikola-rados to have a look since `wps_generate_climos` is implemented in `thunderbird` repo.

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
 cdo = Cdo()
 
 
+def has_climo_periods(input_file, climo):
```
nikola-rados, line 42: This is a bit of a nit-pick but I think the name of this method suggests the output will be a `boolean`.  Something like `get_climo_periods` may be more appropriate.

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+            period,
+            outdir,
+            input_file,
+            operation,
+            *t_range,
+            convert_longitudes=kwargs["convert_longitudes"],
+            split_vars=kwargs["split_vars"],
+            split_intervals=kwargs["split_intervals"],
+            output_resolutions=kwargs["resolutions"]
+        )
```
rod-glover, line 119: I'd guess you are doing this (`kwargs["name"]`) in order to reduce the number of arguments to the function, which is fair enough. However it actually makes the code less clear rather than more clear. Consider the following two options:
- Restore the original named arguments (and don't sweat Code Climate's objections)
- Pass `kwargs` through untouched: `create_climo_files(..., **kwargs)`. You'll need to make sure that `create_climo_files` provides appropriate defaults for unspecified arguments. But that's a good thing ... the defaulting should be done in as few places as possible; avoid repetition unless it actually gives you something.

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+    warn if there is no such variable.
+    '''
+    periods = input_file.climo_periods.keys() & climo
+    logger.info("climo_periods: {}".format(periods))
+    if len(periods) == 0:
+        logger.info(f"{input_file.filepath()} has no variable 'climo_periods'")
+    return periods
+
+
+def is_NetCDF(filepath):
```
nikola-rados, line 55: Two comments here:
1. Again, the name here is suggesting we would get some sort of `boolean` out of it.  `get_cdfdataset` may work better here? I'll leave it up to you.
2. We don't want to have capital letters in our method names unless we are using camelCaps.

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+    '''
+    This function checks if the given input file path is a vaild NetCDF file. It returns
+    NetCDF dataset of the input file. Ohterwise, it halts the program.
+    '''
+    try:
+        return CFDataset(filepath)
+    except Exception as e:
+        logger.info("{}: {}".format(e.__class__.__name__, e))
+        sys.exit()
```
nikola-rados, line 64: This is a fairly aggressive exit strategy but I think it works in the context of a script.

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+
+
+def dry_run_handler(filepath, climo):
+    '''
+    This function handles dry-run operation of generate_climos. The purpose of this function
+    is to check variables and attrbutes to be used for generate_climos. dry-run will not 
+    produca any output files.
```
nikola-rados, line 71: _produce_

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+    NetCDF dataset of the input file. Ohterwise, it halts the program.
+    '''
+    try:
+        return CFDataset(filepath)
+    except Exception as e:
+        logger.info("{}: {}".format(e.__class__.__name__, e))
+        sys.exit()
+
+
+def dry_run_handler(filepath, climo):
```
nikola-rados, line 67: This is great for us in `thunderbird` since we can now directly call this method.  Nice idea here!

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+
+
+def is_NetCDF(filepath):
+    '''
+    This function checks if the given input file path is a vaild NetCDF file. It returns
+    NetCDF dataset of the input file. Ohterwise, it halts the program.
+    '''
+    try:
+        return CFDataset(filepath)
+    except Exception as e:
```
rod-glover, line 62: Good idea to catch this exception and print a log message. However, a few coding points here:
- It is better not to catch a generic `Exception` -- much too broad. Find out what exception is raised  when a non-NetCDF4 file is attempted to be opened, and catch that.
- It should not, generally speaking, be the business of a method called several levels deep to call `sys.exit()`. That should be the business of the highest levels of the script. If the code in question is going to be orchestrated by some higher level function or process (e.g. for parallelization, or some other purpose), `sys.exit()` makes it impossible for the coordinator to do its job.
- This is not an `info` level error: It is a serious error that means the script cannot do anything at all, and should be logged as such. Consider `log.exception()` or `log.critical()` here.

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+    '''
+    This function handles dry-run operation of generate_climos. The purpose of this function
+    is to check variables and attrbutes to be used for generate_climos. dry-run will not 
+    produca any output files.
+    '''
+    logger.info("")
+    logger.info("File: {}".format(filepath))
+    input_file = is_NetCDF(filepath)
+    has_climo_periods(input_file, climo)
+
```
rod-glover, line 77: A general rule of coding is DRY - Don't Repeat Yourself. In this method and in `generate_climos`, you have repeated  the same sequence of code: log some info (the same info), open the file, get the climo periods. This should be factored out into the method that calls these two functions, with arguments as appropriate for the results. (You can see why this was rolled up into a single method before.) 

(There's no benefit to slightly changing what is logged between dry run and full run; in fact it is better if they are as close as possible.)

With this in mind, we see that function `is_NetCDF` becomes unnecessary, and its problems (calling `sys.exit()` too far down the hierarchy) can be fixed by putting its work higher up the calling hierarchy.

The same is probably also true of `has_climo_periods`; that is, it is unnecessary given that it would only be called once.

This refactoring is definitely on the right track, but the repetition these two methods indicates a need to reorganize it a little.

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+        create_climo_files(
+            period,
+            outdir,
+            input_file,
+            operation,
+            *t_range,
+            convert_longitudes=kwargs["convert_longitudes"],
+            split_vars=kwargs["split_vars"],
+            split_intervals=kwargs["split_intervals"],
+            output_resolutions=kwargs["resolutions"]
```
nikola-rados, line 118: What is the behaviour of these variables if the keys you are searching for are not available?

dp/generate_climos.py
rod-glover: Also see note below as to why we probably don't even want or need this function.

tests/test_generate_climos.py
```diff
@@ -53,9 +58,53 @@ def basename_components(filepath):
...
+        outdir,
+        ANY,
+        operation,
+        ANY,
+        ANY,
+        convert_longitudes=convert_longitudes,
+        split_vars=split_vars,
+        split_intervals=split_intervals,
+        output_resolutions=resolutions,
+    )
```
nikola-rados, line 105: Would you be able to explain what is happening here?

dp/generate_climos.py
```diff
@@ -754,7 +834,7 @@ def update_generate_climos_history(args, netCDF_file, time_cdo_format, position=
 
     """
     # command line input through terminal
-    if(len(sys.argv) > 1):
+    if len(sys.argv) > 1:
```
rod-glover, line 837: :+1: horrible C style syntax be gone!

tests/test_generate_climos.py
rod-glover, line 105: :100:

tests/conftest.py
```diff
@@ -8,6 +8,11 @@ def get_dataset(filename):
     return CFDataset(resource_filename(__name__, 'data/tiny_{}.nc').format(filename))
 
 
+@fixture
+def tiny_filepath(request):
+    return resource_filename(__name__, 'data/tiny_{}.nc').format(request.param)
```
rod-glover, line 13: Nice! I wonder if you can DRY up `tiny_dataset` now by using this fixture in it. Not a big deal (a few characters of code), but doing a thing only in one place means you only have to fix it once if you make a mistake.

dp/generate_climos.py
```diff
@@ -39,6 +39,86 @@
...
+def has_climo_periods(input_file, climo):
+    '''
+    This function checks if the given input NetCDF dataset has a variable "climo_periods".
+    generate_climos requires the input dataset to have the variable, thus this function will
+    warn if there is no such variable.
+    '''
+    periods = input_file.climo_periods.keys() & climo
+    logger.info("climo_periods: {}".format(periods))
+    if len(periods) == 0:
+        logger.info(f"{input_file.filepath()} has no variable 'climo_periods'")
```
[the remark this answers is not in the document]
sum1lim: > Will there actually be any output without this?

I believe there is no output if the given input does not have climo periods.

> Also, do we want to continue the script if we have no climo periods? 

@rod-glover what do you think? Should the program halt here?

dp/generate_climos.py
sum1lim: Default values for stdin arguments in [generate_climos](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/scripts/generate_climos#L60) script assure that keys are always available. We might also want to define default arguments in `thunderbird` as well.

dp/generate_climos.py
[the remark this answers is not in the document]
nikola-rados: > I believe there is no output if the given input does not have climo periods.

We can wait to hear what Rod says but I think if there is no output then there is no reason to continue. That's just my $0.02 though.

dp/generate_climos.py
rod-glover: See also comments about DRY above.

tests/test_generate_climos.py
sum1lim: The PR is made because there was a bug that `split_intervals` was not passed into `generate_climos` and was always set to the default value `True`. So, this test is to actually check if every given argument is passed into `create_climo_files` safely. `mock_ccf` mocks the `create_climo_files`, and `assert_called_with` function asserts if the arguments are passed into `mock_ccf` with expected values.

dp/generate_climos.py
jameshiebert: Agreed with point 1.

dp/generate_climos.py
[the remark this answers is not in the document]
rod-glover: You can bail here if you want, or you can just let the logic of the code do its thing and produce nothing more. In general, I favour the early-return style because it tells you where you can stop thinking (always good for me :) ). Not so sure here as bailing is, I think, a call to `sys.exit()`, which is much more problematic than a `return`. However, with the refactoring I suggest, that call to `sys.exit()` may end up at the very top level, which would then be OK.

TL;DR: Let's wait till you refactor and look again. Probably bail, though.

dp/generate_climos.py
rod-glover: See my comments below. I don't favour this.

dp/generate_climos.py
rod-glover: Echoes my thoughts.

dp/generate_climos.py
rod-glover: Definitely don't recommend `kwargs` approach. See my comments below.

dp/generate_climos.py
```diff
@@ -39,17 +39,96 @@
...
+    property. It returns the periods if found, outherwise it halts the program.
+    '''
+    logger.info("")
+    logger.info("Processing: {}".format(filepath))
+
+    try:
+        input_file = CFDataset(filepath)
+    except OSError as e:
+        logger.info("{}: {}".format(e.__class__.__name__, e))
+        return
```
rod-glover, line 57: You are returning nothing (no value) here, but the normal return returns a pair. This will cause the caller some pretty serious problems, since it is going to be expecting a pair. Suggest `return None, []` -- no file, empty period list.

In general, we avoid returning different types of results from the same function. And especially we avoid returning different types when the caller can't tell in advance what to expect, as here.

dp/generate_climos.py
```diff
@@ -39,17 +39,96 @@
...
+
+    periods = input_file.climo_periods.keys() & climo
+    logger.info("climo_periods: {}".format(periods))
+    if len(periods) == 0:
+        logger.critical(f"{input_file.filepath()} has no 'climo_periods' found")
+        return
```
rod-glover, line 63: Ditto regarding return value.

dp/generate_climos.py
```diff
@@ -39,17 +39,96 @@
 cdo = Cdo()
 
 
+def input_check(filepath, climo):
+    '''
+    This function runs general checks on the given input arguments filepath and climo.
+    First, it checks if the given input file path is a vaild NetCDF file. It returns
+    NetCDF dataset of the input file. Ohterwise, it halts the program. Next, it checks
+    if the input NetCDF dataset has a set of standard climo periods by using climo_periods
+    property. It returns the periods if found, outherwise it halts the program.
```
rod-glover, line 48: It doesn't actually halt the program, it just returns (and see comments below about _what_ to return). The behaviour is correct, but the description needs a little tweaking.

Also, we should document what `climo` is: which is, I believe a list of strings naming climo periods.

dp/generate_climos.py
```diff
@@ -39,17 +39,96 @@
...
+    return input_file, periods
+
+
+def dry_run_handler(filepath, climo):
+    '''
+    This function handles dry-run operation of generate_climos. The purpose of this function
+    is to check variables and attrbutes to be used for generate_climos. dry-run will not 
+    produce any output files.
+    '''
+    input_file, periods = input_check(filepath, climo)
```
rod-glover, line 74: Much better! DRY as a bone!

dp/generate_climos.py
```diff
@@ -39,17 +39,96 @@
...
+            period,
+            outdir,
+            input_file,
+            operation,
+            *t_range,
+            convert_longitudes=convert_longitudes,
+            split_vars=split_vars,
+            split_intervals=split_intervals,
+            output_resolutions=resolutions,
+        )
```
rod-glover, line 118: This is very nice. The two methods `dry_run_handler` and `generate_climos` are short, simple, and clear.

tests/conftest.py
```diff
@@ -8,9 +8,18 @@ def get_dataset(filename):
...
+
+@fixture
+def tiny_filepath(request):
+    return get_filepath(request.param)
+
+
 @fixture
 def tiny_dataset(request):
-    return CFDataset(resource_filename(__name__, 'data/tiny_{}.nc').format(request.param))
+    return CFDataset(get_filepath(request.param))
```
rod-glover, line 22: :100:

PR #136, opened 2020-07-17
Flexible logging out strategy for generate_climos dry_run

### LINKED ISSUE #135: `dry_run_handler` needs to have a flexible output
We should style the [`dry_run_handler`](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/dp/generate_climos.py#L72) after the [`dry_run`](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/dp/generate_prsn.py#L22) method from `generate_prsn`. This style would give us more flexibility when using the method in `thunderbird`.

sum1lim:
This PR resolves #135 

For the convenience of `thunderbird`'s `wps_generate_climos` dry_run output, [dry_run_helper](https://github.com/pacificclimate/climate-explorer-data-prep/blob/i135/dp/generate_climos.py#L73) in `generate_climos.py` follows the similar logging strategy of `generate_prsn.py`'s [dry_run](https://github.com/pacificclimate/climate-explorer-data-prep/blob/master/dp/generate_prsn.py#L22)

### DIFF
--- a/dp/generate_climos.py
+++ b/dp/generate_climos.py
@@ -39,8 +39,8 @@ logger.setLevel(logging.DEBUG)  # For testing, overridden by -l when run as a sc
 cdo = Cdo()
 
 
-def input_check(filepath, climo):
-    '''
+def input_check(filepath, climo, outputer=logger.info):
+    """
     This function runs general checks on the given input arguments filepath and climo.
     There are 2 checking processes:
 
@@ -50,9 +50,8 @@ def input_check(filepath, climo):
 
     The function returns CFDataset and list.
     If any of the checks are not passed, the function returns None and empty list.
-    '''
-    logger.info("")
-    logger.info("Processing: {}".format(filepath))
+    """
+    outputer("File: {}".format(filepath))
 
     try:
         input_file = CFDataset(filepath)
@@ -61,34 +60,48 @@ def input_check(filepath, climo):
         return None, []
 
     periods = input_file.climo_periods.keys() & climo
-    logger.info("climo_periods: {}".format(periods))
+    outputer("climo_periods: {}".format(periods))
     if len(periods) == 0:
-        logger.critical(f"{input_file.filepath()} contains no standard climatological periods")
+        logger.critical(
+            f"{input_file.filepath()} contains no standard climatological periods"
+        )
         return None, []
 
     return input_file, periods
 
 
-def dry_run_handler(filepath, climo):
-    '''
+def dry_run_handler(filepath, climo, outpath=None):
+    """
     This function handles dry-run operation of generate_climos. The purpose of this function
-    is to check variables and attrbutes to be used for generate_climos. dry-run will not 
-    produce any output files.
-    '''
-    input_file, periods = input_check(filepath, climo)
+    is to check variables and attrbutes to be used for generate_climos. 
 
-    for attr in ['project', 'institution', 'model', 'emissions', 'run']:
+        * dry-run will not produce any NetCDF output files.
+        * dry-run information will be logged out to an output .txt file if outpath is given
+        * Otherwise, it will be logged out to terminal
+    """
+    if outpath:  # Used for wps process in thunderbird
+        output_items = []
+        outputer = output_items.append
+    else:
+        outputer = logger.info
+
+    outputer("DRY_RUN")
+
+    input_file, periods = input_check(filepath, climo, outputer)
+
+    for attr in ["project", "institution", "model", "emissions", "run"]:
         try:
-            logger.info(
-                "{}: {}".format(attr, getattr(input_file.metadata, attr))
-            )
+            outputer("{}: {}".format(attr, getattr(input_file.metadata, attr)))
         except Exception as e:
-            logger.info("{}: {}: {}".format(attr, e.__class__.__name__, e))
-    logger.info(
-        "dependent_varnames: {}".format(input_file.dependent_varnames())
-    )
-    for attr in ['time_resolution', 'is_multi_year_mean']:
-        logger.info("{}: {}".format(attr, getattr(input_file, attr)))
+            outputer("{}: {}: {}".format(attr, e.__class__.__name__, e))
+    outputer("dependent_varnames: {}".format(input_file.dependent_varnames()))
+    for attr in ["time_resolution", "is_multi_year_mean"]:
+        outputer("{}: {}".format(attr, getattr(input_file, attr)))
+
+    if outpath:
+        with open(outpath, "w") as f:
+            for line in output_items:
+                f.write("{}\n".format(line))
 
 
 def generate_climos(
@@ -101,10 +114,10 @@ def generate_climos(
     split_intervals=True,
     resolutions={"yearly", "seasonal", "monthly"},
 ):
-    '''
+    """
     This function runs general generate_climos operation. The main purpose of this function
     is to call create_climo_files.
-    '''
+    """
     input_file, periods = input_check(filepath, climo)
 
     for period in periods:
--- a/dp/generate_climos.py
+++ b/dp/generate_climos.py
@@ -51,7 +51,7 @@ def input_check(filepath, climo, outputer=logger.info):
     The function returns CFDataset and list.
     If any of the checks are not passed, the function returns None and empty list.
     """
-    outputer("File: {}".format(filepath))
+    outputer(f"File: {filepath}")
 
     try:
         input_file = CFDataset(filepath)
@@ -60,7 +60,7 @@ def input_check(filepath, climo, outputer=logger.info):
         return None, []
 
     periods = input_file.climo_periods.keys() & climo
-    outputer("climo_periods: {}".format(periods))
+    outputer(f"climo_periods: {periods}")
     if len(periods) == 0:
         logger.critical(
             f"{input_file.filepath()} contains no standard climatological periods"
@@ -91,17 +91,17 @@ def dry_run_handler(filepath, climo, outpath=None):
 
     for attr in ["project", "institution", "model", "emissions", "run"]:
         try:
-            outputer("{}: {}".format(attr, getattr(input_file.metadata, attr)))
+            outputer(f"{attr}: {getattr(input_file.metadata, attr)}")
         except Exception as e:
-            outputer("{}: {}: {}".format(attr, e.__class__.__name__, e))
+            outputer(f"{attr}: {e.__class__.__name__}: {e}")
     outputer("dependent_varnames: {}".format(input_file.dependent_varnames()))
     for attr in ["time_resolution", "is_multi_year_mean"]:
-        outputer("{}: {}".format(attr, getattr(input_file, attr)))
+        outputer(f"{attr}: {getattr(input_file, attr)}")
 
     if outpath:
         with open(outpath, "w") as f:
             for line in output_items:
-                f.write("{}\n".format(line))
+                f.write(f"{line}\n")
 
 
 def generate_climos(
--- a/dp/generate_climos.py
+++ b/dp/generate_climos.py
@@ -63,7 +63,7 @@ def input_check(filepath, climo, outputer=logger.info):
     outputer(f"climo_periods: {periods}")
     if len(periods) == 0:
         logger.critical(
-            f"{input_file.filepath()} contains no standard climatological periods"
+            f"{input_file.filepath()} contains no standard climatological periods {climo}"
         )
         return None, []
 
--- a/dp/generate_climos.py
+++ b/dp/generate_climos.py
@@ -94,7 +94,7 @@ def dry_run_handler(filepath, climo, outpath=None):
             outputer(f"{attr}: {getattr(input_file.metadata, attr)}")
         except Exception as e:
             outputer(f"{attr}: {e.__class__.__name__}: {e}")
-    outputer("dependent_varnames: {}".format(input_file.dependent_varnames()))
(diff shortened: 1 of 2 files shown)


dp/generate_climos.py
```diff
@@ -50,9 +50,8 @@ def input_check(filepath, climo):
 
     The function returns CFDataset and list.
     If any of the checks are not passed, the function returns None and empty list.
-    '''
-    logger.info("")
-    logger.info("Processing: {}".format(filepath))
+    """
+    outputer("File: {}".format(filepath))
```
nikola-rados, line 54: Maybe go with an `f-string` here.

dp/generate_climos.py
```diff
@@ -61,34 +60,48 @@ def input_check(filepath, climo):
         return None, []
 
     periods = input_file.climo_periods.keys() & climo
-    logger.info("climo_periods: {}".format(periods))
+    outputer("climo_periods: {}".format(periods))
```
nikola-rados, line 63: Ditto above.

dp/generate_climos.py
```diff
@@ -61,34 +60,48 @@ def input_check(filepath, climo):
...
+            f"{input_file.filepath()} contains no standard climatological periods"
+        )
         return None, []
 
     return input_file, periods
 
 
-def dry_run_handler(filepath, climo):
-    '''
+def dry_run_handler(filepath, climo, outpath=None):
```
nikola-rados, line 73: Above you defaulted this value to `logger.info`, I think we could do that here as well.

scripts/generate_climos
```diff
@@ -10,9 +11,14 @@ from dp.generate_climos import logger, generate_climos, dry_run_handler
...
     if args.dry_run:
-        logger.info("DRY RUN")
         for filepath in args.filepaths:
-            dry_run_handler(filepath, args.climo)
+            filename = filepath.split("/")[-1][:-3]
+            outpath = (
+                os.path.join(args.outdir, filename + "_dry_run.txt")
+                if args.dry_output_to_file
+                else None
+            )
```
nikola-rados, line 20: I think we want to have the default behaviour log to the `python` `logging` module rather than to a file.

scripts/generate_climos
```diff
@@ -59,6 +65,12 @@ if __name__ == "__main__":
         default="INFO",
     )
     parser.add_argument("-n", "--dry-run", dest="dry_run", action="store_true")
+    parser.add_argument(
+        "-f",
+        "--dry-output-to-file",
```
nikola-rados, line 70: We can probably simplify this to `--dry-file`.

scripts/generate_climos
```diff
@@ -89,7 +101,11 @@ if __name__ == "__main__":
         choices=["mean", "std"],
     )
     parser.set_defaults(
-        dry_run=False, convert_longitudes=True, split_vars=True, split_intervals=True
+        dry_run=False,
+        dry_output_to_file=False,
+        convert_longitudes=True,
+        split_vars=True,
+        split_intervals=True,
     )
```
nikola-rados, line 109: I know this was here before, but I would suggest moving these defaults into their definitions. I'll leave this one up to you though, whatever you feel is more clear.

dp/generate_climos.py
```diff
@@ -61,34 +60,48 @@ def input_check(filepath, climo):
...
-            logger.info("{}: {}: {}".format(attr, e.__class__.__name__, e))
-    logger.info(
-        "dependent_varnames: {}".format(input_file.dependent_varnames())
-    )
-    for attr in ['time_resolution', 'is_multi_year_mean']:
-        logger.info("{}: {}".format(attr, getattr(input_file, attr)))
+            outputer("{}: {}: {}".format(attr, e.__class__.__name__, e))
+    outputer("dependent_varnames: {}".format(input_file.dependent_varnames()))
+    for attr in ["time_resolution", "is_multi_year_mean"]:
+        outputer("{}: {}".format(attr, getattr(input_file, attr)))
```
nikola-rados, line 99: `f-string`

dp/generate_climos.py
```diff
@@ -61,34 +60,48 @@ def input_check(filepath, climo):
...
+            outputer("{}: {}".format(attr, getattr(input_file.metadata, attr)))
         except Exception as e:
-            logger.info("{}: {}: {}".format(attr, e.__class__.__name__, e))
-    logger.info(
-        "dependent_varnames: {}".format(input_file.dependent_varnames())
-    )
-    for attr in ['time_resolution', 'is_multi_year_mean']:
-        logger.info("{}: {}".format(attr, getattr(input_file, attr)))
+            outputer("{}: {}: {}".format(attr, e.__class__.__name__, e))
+    outputer("dependent_varnames: {}".format(input_file.dependent_varnames()))
```
nikola-rados, line 97: `f-string`

dp/generate_climos.py
```diff
@@ -61,34 +60,48 @@ def input_check(filepath, climo):
...
+    outputer("DRY_RUN")
+
+    input_file, periods = input_check(filepath, climo, outputer)
+
+    for attr in ["project", "institution", "model", "emissions", "run"]:
         try:
-            logger.info(
-                "{}: {}".format(attr, getattr(input_file.metadata, attr))
-            )
+            outputer("{}: {}".format(attr, getattr(input_file.metadata, attr)))
```
nikola-rados, line 94: `f-string`

dp/generate_climos.py
```diff
@@ -61,34 +60,48 @@ def input_check(filepath, climo):
...
-        logger.info("{}: {}".format(attr, getattr(input_file, attr)))
+            outputer("{}: {}: {}".format(attr, e.__class__.__name__, e))
+    outputer("dependent_varnames: {}".format(input_file.dependent_varnames()))
+    for attr in ["time_resolution", "is_multi_year_mean"]:
+        outputer("{}: {}".format(attr, getattr(input_file, attr)))
+
+    if outpath:
+        with open(outpath, "w") as f:
+            for line in output_items:
+                f.write("{}\n".format(line))
```
nikola-rados, line 104: `f-string`

scripts/generate_climos
sum1lim: I believe it defaults to `logging` module if `-f` argument is not provided because of the `if args.dry_output_to_file else None` statement.

sum1lim:
I think my logic is pretty solid on defaulting and passing `outpath` argument to `dry_run_handler`. If `--dry-file` and `-f` argument is not passed from the user input, [outpath is defaulted to None](https://github.com/pacificclimate/climate-explorer-data-prep/blob/i135/scripts/generate_climos#L19). When `None` value is passed for `outpath`, then the `dry_run_handler` sets the [outputer as logger.info](https://github.com/pacificclimate/climate-explorer-data-prep/blob/i135/dp/generate_climos.py#L86).

dp/generate_climos.py
nikola-rados: Missed this one!

scripts/generate_climos
nikola-rados: You are right it will do that. 

I feel like we can make this a bit more clear. Instead of generating the filename for the output we should have `None` as the default and allow the user to specify names if they so choose. This would mean turning `dry_output_to_file` into a `string` rather than a `boolean`. That way you could just pass that directly into the method without having to process it into `outpath`.

dp/generate_climos.py
```diff
@@ -61,34 +60,48 @@ def input_check(filepath, climo):
         return None, []
 
     periods = input_file.climo_periods.keys() & climo
-    logger.info("climo_periods: {}".format(periods))
+    outputer(f"climo_periods: {periods}")
     if len(periods) == 0:
-        logger.critical(f"{input_file.filepath()} contains no standard climatological periods")
+        logger.critical(
+            f"{input_file.filepath()} contains no standard climatological periods"
+        )
```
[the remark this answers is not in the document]
sum1lim: I am not so sure about what you meant by `our intention of generalizing the output method`. Did you mean it is against the format of `dry_run` logging such as `attribute_key: attribute_value`? If so, I think it is OK the way it is since `input_check` is not only called in `dry_run_handler` but also the general `generate_climos`, and the messages are to tell what problems caused the error, which is different from the purpose of other logging messages in `dry_run_handler`.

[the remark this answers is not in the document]
nikola-rados: Reading back my comment I don't think I made it very clear what my concern is.

We want to use `outputer` to generalize how we are logging/writing the messages. The method shouldn't care what the `outputer` actually is, only that it takes a single argument that is a string and we expect it to be stored/written somewhere. If we are using `logging.info` we can't log messages that are higher levels than that (i.e. `critical`). If we want to use higher levels we need to sort out a generic way of going about it.

Does that clear things up a bit?

scripts/generate_climos
sum1lim: Right, I actually implemented this way, because I thought it needs a static naming convention to be called in `wps_generate_climos`. But after double-checking, it is OK for the output file to have a user-defined name.

sum1lim: @nikola-rados Ahh, there is another reason why I did this. Unlike `generate_prsn`, `generate_climos` can take multiple input files and produce multiple output files as well. Therefore, if a user gives one argument for `-f/--dry-file`, the output file will contain contents only for the last input file generated. For, example, if we run the following command;
```
generate_climos -o output ./tests/data/tiny_daily_pr.nc ./tests/data/tiny_daily_prsn.nc -p mean -n -f dry_run.txt
```
`dry_run.txt` will only have the dry_run info for `./tests/data/tiny_daily_prsn.nc`.

I have two ideas on this problem:
1. keep the static naming method that I used
2. let users give multiple output file names, but the number of them should match the same number of input arguments. Otherwise, throw an error.
```
generate_climos -o output ./tests/data/tiny_daily_pr.nc ./tests/data/tiny_daily_prsn.nc -p mean -n -f pr_dry_run.txt prsn_dry_run.txt
```

dp/generate_climos.py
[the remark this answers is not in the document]
sum1lim: >If we are using logging.info we can't log messages that are higher levels than that (i.e. critical). If we want to use higher levels we need to sort out a generic way of going about it.

@nikola-rados I am still not sure if I am following. Do you want the messages logged with `CRITICAL` level shown with the `dry-run` info messages, which are logged with `INFO` level? If so, I am seeing both messages when I ran the following commands.
```
(venv) slim@devel2:~/Desktop/climate-explorer-data-prep$ generate_climos -o output ./tests/data/tiny_daily_pr.c -p mean -n -f
-->> Could not load xarray!! <<--
2020-07-21 10:56:26 CRITICAL: OSError: No such file or directory
Traceback (most recent call last):
  File "/home/slim/Desktop/climate-explorer-data-prep/venv/bin/generate_climos", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/slim/Desktop/climate-explorer-data-prep/scripts/generate_climos", line 116, in <module>
    main(args)
  File "/home/slim/Desktop/climate-explorer-data-prep/scripts/generate_climos", line 21, in main
    dry_run_handler(filepath, args.climo, outpath)
  File "/home/slim/Desktop/climate-explorer-data-prep/dp/generate_climos.py", line 97, in dry_run_handler
    outputer(f"dependent_varnames: {input_file.dependent_varnames()}")
AttributeError: 'NoneType' object has no attribute 'dependent_varnames'
```
```
(venv) slim@devel2:~/Desktop/climate-explorer-data-prep$ generate_climos -o output ./tests/data/tiny_downscaled_pr.nc -p mean --climo 2080 -n
-->> Could not load xarray!! <<--
2020-07-21 11:08:31 INFO: DRY_RUN
2020-07-21 11:08:31 INFO: File: ./tests/data/tiny_downscaled_pr.nc
2020-07-21 11:08:31 INFO: climo_periods: set()
2020-07-21 11:08:31 CRITICAL: ./tests/data/tiny_downscaled_pr.nc contains no standard climatological periods ['2080']
2020-07-21 11:08:31 INFO: project: AttributeError: 'NoneType' object has no attribute 'metadata'
2020-07-21 11:08:31 INFO: institution: AttributeError: 'NoneType' object has no attribute 'metadata'
2020-07-21 11:08:31 INFO: model: AttributeError: 'NoneType' object has no attribute 'metadata'
2020-07-21 11:08:31 INFO: emissions: AttributeError: 'NoneType' object has no attribute 'metadata'
2020-07-21 11:08:31 INFO: run: AttributeError: 'NoneType' object has no attribute 'metadata'
Traceback (most recent call last):
  File "/home/slim/Desktop/climate-explorer-data-prep/venv/bin/generate_climos", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/slim/Desktop/climate-explorer-data-prep/scripts/generate_climos", line 116, in <module>
    main(args)
  File "/home/slim/Desktop/climate-explorer-data-prep/scripts/generate_climos", line 21, in main
    dry_run_handler(filepath, args.climo, outpath)
  File "/home/slim/Desktop/climate-explorer-data-prep/dp/generate_climos.py", line 97, in dry_run_handler
    outputer(f"dependent_varnames: {input_file.dependent_varnames()}")
AttributeError: 'NoneType' object has no attribute 'dependent_varnames'
```

sum1lim:
After a discussion with @nikola-rados, we have decided not to add functionality to output dry-run info to a file, which means the PR is to be closed. We could not find a necessity to have the functionality after figuring out logging messages could be redirected to a file in `thunderbird` using `logging.basicConfig`.

PR #147, opened 2022-01-14
Update nchelpers

### LINKED ISSUE #146: Update nchelpers to 5.5.8
In order to enable the scripts to handle CMIP6 data, nchelpers must be updated to this version.

eyvorchuk:
This PR resolves #146, which updates `nchelpers` to version 5.5.8 to handle CMIP6 data. This update caused `netcdf4` to be updated from 1.3.0 to 1.5.8 and each test in `test_decompose_flow_vectors.py` failed with the following error:

```bash
tests/test_decompose_flow_vectors.py:29: in create_routing_file
    flows[:] = routes
src/netCDF4/_netCDF4.pyx:4903: in netCDF4._netCDF4.Variable.__setitem__
    ???
../../.local/share/virtualenvs/climate-explorer-data-prep-s1-BAKit/lib/python3.8/site-packages/netCDF4/utils.py:356: in _StartCountStride
    datashape = broadcasted_shape(shape, datashape)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

shp1 = (3, 3), shp2 = (9,)

    def broadcasted_shape(shp1, shp2):
        # determine shape of array of shp1 and shp2 broadcast against one another.
        x = np.array([1])
        # trick to define array with certain shape that doesn't allocate all the
        # memory.
        a = as_strided(x, shape=shp1, strides=[0] * len(shp1))
        b = as_strided(x, shape=shp2, strides=[0] * len(shp2))
>       return np.broadcast(a, b).shape
E       ValueError: shape mismatch: objects cannot be broadcast to a single shape.  Mismatch is between arg 0 with shape (3, 3) and arg 1 with shape (9,).

../../.local/share/virtualenvs/climate-explorer-data-prep-s1-BAKit/lib/python3.8/site-packages/netCDF4/utils.py:964: ValueError

```
This was fixed by converting `flows` and `wrong_flows` into 2D arrays.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,6 +14,7 @@ cdo = "==1.5.3"
 nchelpers = "==5.5.8"
 Pint = "*"
 PyYAML = "*"
+dp = {editable = true, path = "."}
 
 [dev-packages]
 pytest = "*"
@@ -22,7 +23,3 @@ black = "==21.7b0"
 
 [requires]
 python_version = "3.8"
-
-[packages.e1839a8]
-path = "."
-editable = true
--- a/Pipfile
+++ b/Pipfile
@@ -11,7 +11,7 @@ name = "pcic"
 [packages]
 python-dateutil = "*"
 cdo = "==1.5.3"
-nchelpers = "==5.5.6"
+nchelpers = "==5.5.8"
 Pint = "*"
 PyYAML = "*"
 
--- a/tests/test_decompose_flow_vectors.py
+++ b/tests/test_decompose_flow_vectors.py
@@ -26,8 +26,10 @@ def create_routing_file(name, numlats, numlons, routes):
 
     lats[:] = range(45, 60, int(15 / numlats))
     lons[:] = range(-125, -110, int(15 / numlons))
-    flows[:] = routes
-    wrong_flows[:] = [100, -25, 358, 14, 5, 68, -7, -128, 15]
+    flows[:] = np.reshape(routes, (numlats, numlons))
+    wrong_flows[:] = np.reshape(
+        [100, -25, 358, 14, 5, 68, -7, -128, 15], (numlats, numlons)
+    )
 
     testfile.close()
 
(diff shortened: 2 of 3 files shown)


jameshiebert:
Cool, I can take a look at this.

tests/test_decompose_flow_vectors.py
```diff
@@ -26,8 +26,10 @@ def create_routing_file(name, numlats, numlons, routes):
 
     lats[:] = range(45, 60, int(15 / numlats))
     lons[:] = range(-125, -110, int(15 / numlons))
-    flows[:] = routes
-    wrong_flows[:] = [100, -25, 358, 14, 5, 68, -7, -128, 15]
+    flows[:] = np.reshape(routes, (numlats, numlons))
```
jameshiebert, line 29: Ah I see... the new version of `netcdf4` is more strict with its broadcasting rules. Your solution looks great.

Pipfile
```diff
@@ -11,9 +11,10 @@ name = "pcic"
 [packages]
 python-dateutil = "*"
 cdo = "==1.5.3"
-nchelpers = "==5.5.6"
+nchelpers = "==5.5.8"
 Pint = "*"
 PyYAML = "*"
+dp = {editable = true, path = "."}
```
jameshiebert, line 17: I'm still not confident that specifying the package under testing within its own Pipfile is the right thing to do. But if it works, then barring better guidance I'm OK with it. We should read up further on `pipenv` and clarify whether this is an intended use case.

jameshiebert (approved):
Looks good to me. Merge away :) And if you're going to use *this version* of the code to do the processing that you have coming up, I would recommend rolling a new tagged release, so that you can definitively say "I used this version of the code to do the CMIP6 processing".

PR #148, opened 2023-12-12
Use Poetry for dependency management

jameshiebert:
## Changes
This PR removes the use of Pipenv for dependency management and replaces it with Python poetry
- `setup.py` has been migrated to `pyproject.toml`
- `pytest.cfg` has been incorporated into `pyproject.toml`
- All GH actions have been migrated to poetry
- Minor modifications were made to the command line scripts to be able to be incorporated into `pyproject.toml` (which requires a module and function as an entry point rather than a script).


.github/workflows/lint.yml
```diff
@@ -5,11 +5,11 @@ on: push
...
   black:
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
```
rod-glover, line 12: The existence of v4 is worth mentioning to all of us in Zulip or at the next team meeting. Are there any noticeable differences?

.github/workflows/pypi-publish.yml
```diff
@@ -8,24 +8,29 @@ on:
...
       env:
         TWINE_USERNAME: ${{ secrets.pcic_at_pypi_username }}
         TWINE_PASSWORD: ${{ secrets.pcic_at_pypi_password }}
       run: |
-        python setup.py sdist bdist_wheel
-        twine upload --repository-url https://pypi.pacificclimate.org/ --skip-existing -u $TWINE_USERNAME -p $TWINE_PASSWORD dist/*
+        # Configure Poetry to publish to PCIC private package repository
+        # Private repo name is "pcic". We could factor that out as an env var...
+        poetry config repositories.pcic https://pypi.pacificclimate.org/
+        poetry config http-basic.pcic $PCIC_PYPI_USERNAME $PCIC_PYPI_PASSWORD
```
rod-glover, line 32: Env var names need to be changed to these ones.

README.md
```diff
@@ -26,18 +26,23 @@ _before_ installing the Python modules:
 ```bash
 $ module load netcdf-bin
 $ module load cdo-bin
+$ module load poetry
 ```
 
-Python installation should be done in a virtual environment. We recommend `pipenv`:
+Python installation should be done in a virtual environment managed by the `poetry` tool:
```
rod-glover, line 32: Suggest adding hyperlink to Poetry documentation.

README.md
```diff
@@ -53,7 +58,7 @@ tests in Travis.
 We can and MUST, however, run the tests on our development machines. It's as simple as
 
 ```bash
-pytest
+poetry run py.test
 ```
 
 (Oh, the unintended punning!)
```
rod-glover, line 64: The punning is less obvious now!

pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[tool.poetry]
+name = "dp"
+version = "0.8.8"
+description = "PCIC Climate Explorer Data Preparation"
+authors = ["Rod Glover <rglover@uvic.ca>"]
```
rod-glover, line 5: More authors?

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+generate_climos = "dp.scripts.generate_climos:runme"
+generate_prsn = "dp.scripts.generate_prsn:runme"
+split_merged_climos = "dp.scripts.split_merged_climos:runme"
+update_metadata = "dp.scripts.update_metadata:runme"
+decompose_flow_vectors = "dp.scripts.decompose_flow_vectors:runme"
+
+[tool.pytest.ini_options]
+markers = [
+    "slow: marks tests that are slow (deselect with '-m \"not slow\"')",
+]
```
rod-glover, line 46: Yay, looks great.

.github/workflows/lint.yml
```diff
@@ -5,11 +5,11 @@ on: push
...
       - uses: psf/black@stable
```
rod-glover, line 13: I've found it better to install Black (with a specific version) as a dev dependency on and invoke it using Python in the CI and from the command line or IDE when developing. Otherwise different versions of Black disagree on their defaults (`@stable` does change), and it is very frustrating. Further, different repos following this patter use different versions, so it is not possible at present to have one global version on your machine.

rod-glover (commented):
Looks great! A few comments inline, and the following:

`pytest`, not `py.test`, has long been the [recommended](https://github.com/pytest-dev/pytest/issues/1629) entry point for Pytest. This occurs in several places in this codebase. Suggest updating throughout.

README.md contains some outdated stuff, including
- mention of TravisCI
- directions in Releasing section involving setup.py; needs updating to pyproject.toml

.github/workflows/lint.yml
jameshiebert: I didn't research any differences... just saw a deprecation notice w.r.t. v2 so I upgraded to the most recent version. It's probably worth mentioning to folks, even if I'm not very well informed about it aside from "it works!".

.github/workflows/lint.yml
jameshiebert: Can you point to a similar config in another repo that we could adopt in this one? (I'm not finding a lint.yml in any other repositories).

My approach to using `black` has essentially been asking it to fix our formatting for us, and its version hasn't been of any consequence.

rod-glover: It is possible that Black has settled down lately, but in the past it changed its default settings from release to release. Which caused angst when it changed under you -- potentially hundreds of changes in a formerly Black-compliant codebase. Hence version pinning. And making it part of the dev deps, so that everyone working on the repo, plus the workflow, is using the same Black, always.

I've noticed that Black checks are often the last step of the Python CI workflow. For example, [in pycds](https://github.com/pacificclimate/pycds/blob/master/.github/workflows/python-ci.yml). I believe this is because a failing Black check can cause your unit tests to be cancelled, and this can be undesirable. I think I recall discussing this with Nik a while back when we were both working on the same repo. In any case, it is common practice in our projects.

jameshiebert:
Going to merge this now, but will file an issue w.r.t. the reviewer's comments about specifying a version of `black` rather than leaving it at "stable".

.github/workflows/lint.yml
jameshiebert: I'd argue that it's a good thing for black checks to be done *first* since they are the cheapest and that we consider them to be important for controlling code quality.

If there is a specific version of black that we should pin to, please feel free to recommend it :)

